PR #1, opened 2016-07-29
Secured Preferences

joeygrover:
To quote the introduction:

> This proposal will be to introduce a closed-source library that provides a secure method for storing information using Android's SharedPreferences system. The library will be owned and maintained by the SDL Maintainers and interested parties from the governing body. 

The original idea was discussed back in September/October of 2015. We decided there was a need to increase security over the standard SharedPreference model. The library was mostly finished later that year. It only recently was integrated and tested with a corresponding pull request.

I know a pain point will be the closed-source status and the license that will be used. I believe it is necessary on both points. The closed-source part will also be available to all maintainers and the governing body.

See my [pull request](https://github.com/smartdevicelink/sdl_android/pull/325) in the sdl_android project with the changes mentioned.

### DIFF
--- a/proposals/NNNN-SecuredPreferences.md
+++ b/proposals/NNNN-SecuredPreferences.md
@@ -8,7 +8,7 @@
 
 ## Introduction
 
-This proposal will be to introduce a closed-source library that is owned and maintained by the SDL Maintainers and interested parties from the governing body.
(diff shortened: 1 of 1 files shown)


joeygrover:
Please leave feedback in issue #3

PR #7, opened 2016-08-11
iOS Stringly Typed Enums

joeljfischer:
This feature is a major change affecting all SDL enums in the iOS library. SDL enums are essentially string enums, but they are not handled as such. We should be handling string enums in the way Apple recommends and in the way that allows for the most Swift compatibility.


joeljfischer:
This should not be made from the master branch of the fork.

PR #9, opened 2016-08-15
iOS RPC Storage Properties as Immutable

joeljfischer:
This feature is a major change affecting RPC requests, responses, and structs. Currently, RPCs largely use mutable dictionaries and arrays. This change modifies all RPC public mutable storage APIs of these types to be immutable instead.


asm09fsu:
This looks fine to me. Relatively unrelated, but still questionable. Shouldn't the properties in `SDLDisplayCapabilities` be `nonatomic`? Perhaps that's for a different proposal. Also, do we need to include both `init`s since it is inheriting them from `SDLRPCStruct`? Again, probably also just a new proposal.

joeljfischer:
@asm09fsu Since we're inheriting them from `SDLRPCStruct`, removing them isn't a change at all and can be done in any hotfix.

PR #10, opened 2016-08-16
iOS RPC Properties as Readonly

joeljfischer:
This feature is a major change affecting RPC requests, responses, and structs. Currently, RPCs use readwrite properties that allow properties to be altered after the object has been created. This proposal will change that such that properties can no longer be set after an object has been created.


joeljfischer:
I'm retracting this PR after feedback, a modified version will be resubmitted at a later time.

PR #14, opened 2016-08-17
Enforce iOS ObjC Generics

joeljfischer:
This feature is a major change affecting primarily RPC requests, responses, and structs, but all classes in the project will be audited and updated. We have many public areas where SDL exposes a collection class; generics would allow us to specify what those collection classes contain.


asm09fsu:
LGTM 👍

PR #38, opened 2016-09-16
iOS SDLJingle Constants Should Be Constants

joeljfischer:
This proposal is a major change to alter `SDLJingle` string constants to be actual string constants.


joeljfischer:
This is being addressed as part of the `SDLEnum` changes.

PR #47, opened 2016-10-04
iOS Library Use Nullability Annotations

joeljfischer:
This proposal is to make major changes to the iOS library to support using nullability annotations in every file. After this proposal is implemented, every file in the project should have nullability information. In addition, we should make improvements in order to have as few `nullable` properties as possible.


joeljfischer:
This is a dupe...

PR #56, opened 2016-11-17
Streaming Media Manager 5.0

asm09fsu:
This proposal is geared at revamping how the Streaming Media Manager (SMM) class manages audio/video streaming. This revamp will also include lifecycle management for these streams as the overall app/proxy lifecycle changes.

### DIFF
--- a/proposals/nnnn-streaming-media-manager-update.md
+++ b/proposals/nnnn-streaming-media-manager-update.md
@@ -33,6 +33,9 @@ There are multiple parts of the proxy lifecycle that the SMM would have an impac
 #### `SDLLifecycleManager`
 If the application has an `SDLLifecycleConfiguration` with the `appType` of `NAVIGATION`, then SMM will be started automatically once `SDLLifecycleManager` enters the state `didEnterStateSettingUpManagers`. This would simply start the manager, not the sessions. At this point, SMM is ready and listening to HMI status changes and will internally start/stop protocols.
 #### `SDLStreamingMediaManager`
+Due to the need to make sure we properly unit test all possible combinations of HMI, app and streaming states, we need to be able to change the state machine at will. This, however, should not be exposed to the developer. Because of this, `SDLStreamingMediaManager` will be a wrapper around `SDLStreamingMediaLifecycleManager`, similar to how `SDLManager` is a wrapper around `SDLLifecycleManager`.
+
+#### `SDLStreamingMediaLifecycleManager`
 ##### State Machines
 There will need to be multiple state machines that need to work together to deal with all the possible scenarios the app can be in on both the phone and head unit. Luckily, HMI states are already provided so we will not be building a new state machine, but will list the possible scenarios for different states below.
 ###### Phone (new state machine)
--- a/proposals/nnnn-streaming-media-manager-update.md
+++ b/proposals/nnnn-streaming-media-manager-update.md
@@ -52,7 +52,7 @@ There will need to be multiple state machines that need to work together to deal
 3. Limited
 	- App was open, however the user has moved to another screen on the head unit. This app is still the primary streaming app. The Streaming State must not move from **Ready**, however if the Phone State is moving from **Active** to **Resigning Active**, we must send 30 frames.  
 4. Full
-	- App is open, and is currently in focus. If the Streaming State is currently **Stopped**, we should proceed to **Starting**, so long as the Phone State is not **Resigning Active** or **Backgrounded**. If moving to this state and the Phone State is currently **Regaining Active** or **Active**, and the Streaming State is currently  **Ready**, the Streaming State should be moved to **Stopped** followed by **Starting** (restarting the streams).
+	- App is open, and is currently in focus. If the Streaming State is currently **Stopped**, we should proceed to **Starting**, so long as the Phone State is not **Resigning Active** or **Backgrounded**. If moving to this state and the Phone State is currently **Regaining Active** or **Active**, and the Streaming State is currently  **Ready**, the Streaming State should be moved to **Stopped** followed by **Starting** (restarting the streams). If moving to this state and the Phone State is currently **Backgrounded**, send an RPC to alert the user to open the app (current suggestion is an Alert with TTS).
 
 ###### Streaming (new state machine)
 1. Stopped
--- a/proposals/nnnn-streaming-media-manager-update.md
+++ b/proposals/nnnn-streaming-media-manager-update.md
@@ -15,8 +15,8 @@ We currently have a SMM which provides the basic functionality for streaming aud
 ## Proposed solution
 #### Current Supported Features:
 - Starting/Stopping video session with different encryption flags
-* Starting/Stopping audio session with different encryption flags
-* Sending video data via `CVImageBufferRef`
+- Starting/Stopping audio session with different encryption flags
+- Sending video data via `CVImageBufferRef`
 - Sending audio data via `NSData` of PCM data
 - Ability to interact with `SDLTouchManager`
 - Ability to set custom video encoder settings
@@ -25,7 +25,8 @@ We currently have a SMM which provides the basic functionality for streaming aud
 
 #### New Features
 - Management of the video session lifecycle in conjunction with proxy & app state changes
-* Management of the audio session lifecycle in conjunction with proxy & app state changes
+- Management of the audio session lifecycle in conjunction with proxy & app state changes
+- Ability to register to a touch manager handler for touch events.
 
 ## Detailed design
 There are multiple parts of the proxy lifecycle that the SMM would have an impact on.
@@ -120,6 +121,21 @@ Since audio data is sent sporadically, we will have a function that will send au
 ```objc
 // Sending Audio Data
 [streamingMediaManager sendAudioData:<#audioData#>];
+```
+
+#### `SDLTouchManager`
+Currently, the `SDLTouchManager` class allows for developers to use a provided set of callbacks so they do not have to provide logic for determining gesture events such as tap, double tap, pinch, and pan. There are some developers, however, that would still like the ability to use their own. Currently, developers can listen to `SDLOnTouchEvent` notifications, and consume the object themselves, however a proposed solution is to make the internal `SDLTouch` object public, and allow for developers to use a handler on `SDLTouchManager` that will provide this touch object back, along with the `SDLTouchType`, to make it easier to handle and consume `SDLOnTouchEvent`.
+
+We would add a property to `SDLTouchManager` called `touchEventHandler`:
+```objc
+typedef void(^SDLTouchEventHandler)(SDLTouch *touch, SDLTouchType type);
+
+/**
+*  @abstract
+*      Returns all OnTouchEvent notifications as SDLTouch and SDLTouchType objects.
+*/
+@property (nonatomic, weak, nullable) SDLTouchEventHandler touchEventHandler;
+
 ```
 
 ## Impact on existing code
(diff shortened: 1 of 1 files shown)


henry-fan:
There is one case which cannot be covered based on our observation. Since iOS10 can auto-detect AppLink apps, there is one scenario where app gets registered and then move to HMI_FULL on user selection, at the meantime app remains background (app enters state machine with initial state Backgrounded), we couldn't do anything about it and user gets a constant "waiting device" screen. The change should also be able to handle this case

asm09fsu:
@henry-fan you are right that is a use case that needs to be handled. What would you expect to see in this use case, as the app is backgrounded and the encoder cannot be used? I am thinking an Alert with tts.

henry-fan:
@asm09fsu it might be the only choice we have.

proposals/nnnn-streaming-media-manager-update.md
```diff
@@ -63,23 +63,23 @@ There will need to be multiple state machines that need to work together to deal
 	- Streaming sessions and encoder are ready. This state is only active so long as HMI is **Limited or greater** and Phone State is **Regaining Active** or **Active**. If at this state, and Phone State changes to **Resigning Active**, send ~30 frames. If at this state, and HMI changes to **Background or less**, move to **Stopped**.
 
 ##### Sending Video Data
-There will be a handler, `videoDataHandler`, that will be within SMM that will expect a return of a `CVImageBufferRef` for video.
```
asm09fsu: @kshala-ford How does this look? Since we are only assuming the developer is using OpenGL, I have changed how we get the video frames from a developer.

kshala-ford: lgtm

kshala-ford:
Unsure if this is covered. What happens to the stream when constantly receiving push notifications? I can imagine scenarios receiving lots of notifications coming from all my facebook/WhatsApp groups. Is it planned to continue the video streaming in that situation?

asm09fsu:
@kshala-ford On iOS currently, notifications show up in a non-evasive manner (banners at the top). This does not change the state of the current foreground app. If an application displays a `UIAlertView` or `UIAlertController`, the app will receive a `UIApplicationWillResignNotification` notification, which will be handled in a similar manner as is described in the doc (sending some frames to the HU or an audio alert, or something.

proposals/nnnn-streaming-media-manager-update.md
```diff
@@ -0,0 +1,148 @@
...
+* Author: [Alex Muller](https://github.com/asm09fsu)
+* Status: **Awaiting review**
+* Impacted Platforms: iOS / Protocol
+
+## Introduction
+This proposal is geared at revamping how the Streaming Media Manager (SMM) class manages audio/video streaming. This revamp will also include lifecycle management for these streams as the overall app/proxy lifecycle changes.
+
+
+## Motivation
+We currently have a SMM which provides the basic functionality for streaming audio and video, however much of the management of this class is still up to the developer, and may vary by implementations. In order to allow developers to deliver high quality mobile navigation experiences, the SMM should be revamped to remove much of the bloat and management of the sessions internally and allow the developers to easily use this class for mobile navigation applications. If you need a real reason for why this should be considered, look at the Detailed Design section for all the possible complexities we expect every developer to implement.
```
joeljfischer: > If you need a real reason for why this should be considered, look at the Detailed Design section for all the possible complexities we expect every developer to implement.

Are you currently giving me the fake reason? 🤔

proposals/nnnn-streaming-media-manager-update.md
```diff
@@ -0,0 +1,148 @@
...
+#### `SDLStreamingMediaManager`
+Due to the need to make sure we properly unit test all possible combinations of HMI, app and streaming states, we need to be able to change the state machine at will. This, however, should not be exposed to the developer. Because of this, `SDLStreamingMediaManager` will be a wrapper around `SDLStreamingMediaLifecycleManager`, similar to how `SDLManager` is a wrapper around `SDLLifecycleManager`.
+
+#### `SDLStreamingMediaLifecycleManager`
+##### State Machines
+There will need to be multiple state machines that need to work together to deal with all the possible scenarios the app can be in on both the phone and head unit. Luckily, HMI states are already provided so we will not be building a new state machine, but will list the possible scenarios for different states below.
+###### Phone (new state machine)
+1. Resigning Active
+	- The app is moving to the background. If the streaming state is **Ready**, and in HMI **Limited or greater**, we should send ~30 frames so that we can alert the user to reopen the app. The streams will remain open. Can move to **Backgrounded**.
```
joeljfischer: Why 30 frames?

proposals/nnnn-streaming-media-manager-update.md
```diff
@@ -0,0 +1,148 @@
...
+2. Backgrounded
+	- The app is currently in the background, and cannot use the encoder to send over frames. Regardless of HMI changes, we cannot stream video. The stream should remain open if currently open. If the HMI changes to **Background or less**, we should close the sessions, and the Streaming state will move to **Stopped**. Can move to **Regaining Active**.
+3. Regaining Active
+	- The app is currently regaining it’s position as the foreground app. If we are currently in HMI **Limited or greater**, and the Streaming state is **Ready**, we will restart the streams by moving to state **Stopped**, and then **Starting**, followed by **Ready**. Can move to **Active**.
```
joeljfischer: Wouldn't this just be that the streaming state machine would just move to **starting**? It should already be in **stopped**, and **ready** is conditional on the streaming state machine, not this.

proposals/nnnn-streaming-media-manager-update.md
```diff
@@ -0,0 +1,148 @@
...
+4. Active
+	- The app is in the foreground. We are able to stream if needed, permitting we are in the proper HMI. Can move to **Resigning Active**.
```
joeljfischer: > permitting

"assuming"?

proposals/nnnn-streaming-media-manager-update.md
```diff
@@ -0,0 +1,148 @@
...
+
+HMI State   | Can Open Audio Stream | Should Close Audio Stream
+------------|-----------------------|--------------------------
+NONE        | No                    | Yes
+BACKGROUND  | No                    | Yes
+LIMITED     | Yes                   | No
+FULL        | Yes                   | No
+
+##### Sending Video Data
+The way in which video frames are sent over will not change from the current implementation. This function will require input of a `CVImageBufferRef` for video. It will be up to the developer to decide how to frequently to call this function. The developer will be able to know if the currently connected stream is encrypted with a `BOOL` `isEncrypted`, a `CGSize` `screenSize` to let them know the current Head Unit screen size, and a `CVPixelBufferPoolRef` for using the encoder’s pixel buffer pool, as is currently available in SMM.
```
joeljfischer: I assume `isEncrypted` et. al. are actually `encrypted` with the getter prepending the `is`?

proposals/nnnn-streaming-media-manager-update.md
```diff
@@ -0,0 +1,148 @@
...
+// Sending Video Data
+CVPixelBufferPoolRef pixelBufferPool = streamingMediaManager.pixelBufferPool;
+CGSize screenSize = streamingMediaManager.screenSize;
+
+CVImageBufferRef frame = <#acquire video frame#>
+[streamingMediaManager sendVideoFrame:frame];
+```
+
+##### Sending video data when iPhone is moving to background
+Since iOS cannot use the encoder when the phone is in the background, to conserve battery and CPU, we must find a solution that will drive the user to reopen the app. The suggested solution is to create a single frame that is generated when SMM’s encoder is initialized, and send that over as the Phone State changes from **Active** to **Resigning Active**. This should be built as a non-customizable screen, so that we can rely on consistent behavior among streaming applications for the time being. The current idea is to have a black screen with the text “Please re-open <#App Name#>”. We have the option to provide translations for all languages that SDL supports, if we wish to go that route. We can also use the app icon, if we wish instead of the app name, as the app name could potentially be quite long.
```
joeljfischer: It's not 100% clear, but it looks like you're saying this would be built into the SMM?

joeljfischer: I would say both name and icon could be helpful. Do we have an idea for how this might be implemented?

proposals/nnnn-streaming-media-manager-update.md
```diff
@@ -0,0 +1,148 @@
...
+
+##### Sending Audio Data
+Since audio data is sent sporadically, we will have a function that will send audio data only when necessary.
```
joeljfischer: Doesn't this already exist? I'm confused if this is a comprehensive overview or just showing the changes?

proposals/nnnn-streaming-media-manager-update.md
```diff
@@ -0,0 +1,148 @@
...
+/**
+*  @abstract
+*      Returns all OnTouchEvent notifications as SDLTouch and SDLTouchType objects.
+*/
+@property (copy, nonatomic, nullable) SDLTouchEventHandler touchEventHandler;
+
+```
+
+## Impact on existing code
+The impact on SDL is that the previous version of `SDLStreamingMediaManager` and all code associated will no longer work. This will require a developer to adapt to using this new method of streaming, but will greatly improve the efficiency of Mobile Navigation Apps.
```
joeljfischer: This seems *mostly* additive, public API-wise. What behavior prevents current code from working with this change?

asm09fsu: This is not only additive. We are completely removing the ability to start/stop the sessions, because we want consistent lifecycle between all streaming apps.

proposals/nnnn-streaming-media-manager-update.md
asm09fsu: This is an entire rewrite of SMM, so this will contain an overview.

proposals/nnnn-streaming-media-manager-update.md
asm09fsu: This _could_ be implemented in SMM, but that's the point of the proposal is to see if that is ideal. We could do an XIB, or dynamically generate an image. Just depends on if we want to give the developers to ability to customize. The only problem with customization is we must make sure it is NHTSA compliant for US market.

proposals/nnnn-streaming-media-manager-update.md
asm09fsu: yes

proposals/nnnn-streaming-media-manager-update.md
asm09fsu: No, as we must stop and restart the sessions by sending an end session and then a start session.

proposals/nnnn-streaming-media-manager-update.md
asm09fsu: 30 frames guarantees that we send enough to be visible on the head unit. This can be changed, but in testing this seems to be a good amount.

proposals/nnnn-streaming-media-manager-update.md
asm09fsu: Yes

proposals/nnnn-streaming-media-manager-update.md
joeljfischer: Then why does `2. Backgrounded` say it's moved to stopped? Is it moving from stopped to stopped again?

asm09fsu: The streaming state is independent from the app and him state. The streaming state moves to stopped only if the HMI state goes to background or none. In all other instances, the stream stays open. Not sure where you are getting stopped -> stopped?

BrandonHe:
@asm09fsu 
I am trying to understand the state switching. Too much states matrix from HMI and App to understand, could you add State Machine Diagram(s) to help us more clearly understand this proposal? Thanks!

PR #61, opened 2016-12-20
Expand Mobile putfile RPC

lchen139:
This proposal expands putfile RPC to incorporate with unreliable wireless medium(like Bluetooth). Introduced CRC32 will add additional data integrity protection for error sensitive scenario ( transfer large binary file).

### DIFF
--- a/nnnn-Expand-Mobile-putfile-RPC.md
+++ b/nnnn-Expand-Mobile-putfile-RPC.md
@@ -1,4 +1,4 @@
-```
+
 # Expand Mobile putfile RPC
 
 * Proposal: [SDL-NNNN](NNNN-filename.md)
@@ -24,6 +24,7 @@ Proposed solution is introducing an optional 4-byte CRC checksum in putfile RPC.
 ## Detailed design
 
 Addition to MOBILE API:
+```
 <enum name="Result" internal_scope="base">
     <element name="Corrupted_DATA">
       <description>The data sent failed to pass CRC check in receiver end</description>
@@ -48,11 +49,11 @@ Addition to MOBILE API:
     </param>
 
   </function> 
-  
+```
+
   
 ## Impact on existing code
 
 
 ## Alternatives considered
 
-````
--- a/nnnn-Expand-Mobile-putfile-RPC.md
+++ b/nnnn-Expand-Mobile-putfile-RPC.md
@@ -55,8 +55,8 @@ Core should not delete file when one chunk was corrupted and it will response "C
 App will know it is corrupted chunk of data, and it will start retransmission for that chunk.
 
 Major change:
-	1. New parameters to putfile request and response 
-    2. Send "Corrupted_DATA" instead of "Invalid_DATA" to app.(SDL core only change)
+		1. New parameters to putfile request and response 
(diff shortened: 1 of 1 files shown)


mfoedisch:
A base assumption for RPC communication was that it is on top of a reliable protocol. If we are using now means of transport that do not follow that assumption, we might be bigger trouble here, e.g. why would this only effect PutFile? What about all the other RPC commands with their JSON and/or binary parts that might get corrupted?

lchen139:
@mfoedisch  thanks your comments. An alternative design in this proposal is about change SDL protocol to add additional robustness.

tpulatha:
@mfoedisch it is correct that we assujme that RPC communication is only done on reliable protocols. Unfortunately, in our experience of implementing SDL we found multiple problems with drivers of embedded systems. These drivers would drop packages or corrupt data packages, this usually leads to smaller issues (like an ignored show). The worst we've seen are app crashes and uploads being corrupted. The nature of very large uploads is that they are very susceptible to corrupted data due to high workload on the drivers. With the current design you can only check a CRC of a, eg, 1GB file after the whole file was transferred, which is a bit late, thus this proposal.

lchen139:
@tpulatha @joeljfischer 
This proposal is ready for review.

mfoedisch:
I'm still a bit confused about the issue we want to solve.

So, the claim is that large file transfers cause transmission errors.

If we, however, transfer the large file in one big chunk and there were errors in the transmission we'd have to re-transmit the whole thing - without guarantee that it would work the next time. (this was @tpulatha point, I believe)

Therefore, we could split a large file in smaller files in order to detect an issue earlier. But this would mean that the issue can also occur with smaller transmission sizes?!

If we see the transmission issues also with smaller data, then the proposed solution would not be sufficient because we have potentially similar sized  transmissions with other RPC commands as well, e.g. RegisterAppInterface response, CreateInteractionChoiceSet, etc.

Considering this, I think, the alternate approach of expanding the SDL protocol instead should be preferred. 

On a different thought, it would be best to fix the driver issues in order to make a platform behave as it is supposed to. Not sure if a high-level solution with all its overhead (file splitting, CRC checking, re-transmit, ...) is the correct thing to do to solve a low-level problem.

A final comment: adding a CRC32 field to PutFile RPC does not hurt - I'm not opposed to that. I just don't think it will solve the underlying issues (as i understand it).

lchen139:
@mfoedisch 

Thanks for your comments.

To clarify,
the issue we reported in this proposal is unreliable medium will have errors during transmission. It will impact every RPC. Among all operations, large file transfer through putfile RPC is most sensitive to this kind of error.

Proposed two solutions: A. "expand putfile" only works as a hot patch to fix putfile issue, B. "expand frame header" protects whole RPC. Only reason I put solution A above solution B is about framework impact. Solution B requires a major change in SDL protocol level. If everyone agrees outcome of solution B worth the change, we can alter the solution sequence. Or, we can implement both solution at a same time for multiple layers of protection.

I agree that it will be perfect to have the issue fixed in driver level. However, we still want to make sure we can maintain functionality while someone is working on to fix driver issue/ driver fix is not possible (in this case, bluetooth can't provide a fix for this issue). The  whole proposal is not a solution. It is about extra layer of robustness protection.

PR #62, opened 2017-01-17
Communication app activation during active embedded audio source or navigation

KhrystynaDubovyk:
Index on Communication app activation during active embedded audio source or navigation was added.

### DIFF
--- a/proposals/0023-Communication-app-activation.md
+++ b/proposals/0023-Communication-app-activation.md
@@ -1,7 +1,7 @@
 # Communication app activation during active embedded audio source or navigation
 
 * Proposal: [SDL-0023]
-* Author: Luxoft/FORD
+* Author: [Melnyk Tetiana](https://github.com/TMelnyk)
 * Status: Awaiting review
 * Impacted Platforms: Core
 [SDL-0023]: https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0023-Communication-app-activation.md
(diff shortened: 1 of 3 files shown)


KhrystynaDubovyk:
@LuxoftAKutsan, @mghiumiusliu, @tmelnyk, @drachenkoanastasiia
please review

asm09fsu:
@KhrystynaDubovyk this needs to be updated to have a more readable name (don't just put the file name).

KhrystynaDubovyk:
hello, @asm09fsu 
could you please specify how I can rename this? is this name meaningless?
> don't just put the file name
what is appropriate to put instead?

joeljfischer:
The title of the PR should be the title of the evolution proposal. In this case: "Communication app activation during active embedded audio source or navigation"

KhrystynaDubovyk:
@ joeljfischer 
should the name of the PR and the title of the evolution proposal be the same?
**ProTip!** says:

> Great commit summaries contain fewer than 50 characters. Place extra information in the extended description 

In this case, shall I rename the title of the evolution proposal to 'Communication app activation'?

joeljfischer:
No, just make the entire name of the PR the same as the title of the evolution proposal. You can ignore the tip.

TMelnyk:
1. Agree with Joe regarding naming
2. This proposal is to provide communication app activation during active embedded sources.
This proposal is to **handle** communication app activation during active **embedded audio and navigation sources.**
3. Motivation  - it is not covered cases actually
But if we`ll rework it - left w/o changes
4. Could you please add some numeric list (1 2 3) to each req at proposed solution
5. Please also add lable or status or smth -  "under_clarification"

TMelnyk:
reviewed

AGaliuzov:
@joeljfischer this one is ready for review

PR #72, opened 2017-01-19
Couple of proposals for improving of SDL internals

mghiumiusliu:
Following parts of SDL requires improvement
- Technology advance
- Architecture

Topics are quite large, detailing will follow.
Additional moments will be disclosed during review

### DIFF
--- a/proposals/nnnn-sdl-environment-and-technology-update.md
+++ b/proposals/nnnn-sdl-environment-and-technology-update.md
@@ -1,0 +1,30 @@
+# Feature name
+
+* Proposal: [SDL-NNNN](nnnn-environment-and-technology-update.md)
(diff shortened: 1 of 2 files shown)


joeljfischer:
Hi @mghiumiusliu, a few notes. First, submitting pull requests, please limit each PR to a single proposal. 

Second, some of these seem too broad. An evolution proposal should be for a single topic of discussion. So, for example, "SDL Refactor" is far too broad. If one piece of this refactor is not accepted, the entire proposal will be rejected. 

Third, please only submit a PR for your suggested updates when you have a full draft or nearly so of your proposal (enough to engender discussion). If you just want to suggest a topic, please use the Slack #sdl_evolution channel instead.

joeljfischer:
Please open separate pull requests for more specific topics.

PR #77, opened 2017-01-20
Providing AppIcon in case of app registration

DrachenkoAnastasiia:

PR #114, opened 2017-03-16
Appicon Resumption

mrapitis:
This proposal is to simplify (and speed up) the process of setting an app icon across app registration requests.


joeljfischer:
This seems very similar to #77, can that be closed?

mrapitis:
Yes, I have asked the requester to close #77 in favor of this PR.

PR #115, opened 2017-03-16
Transfer RPC's with invalid image to HMI

mrapitis:
This proposal provides SDL with the capability to transfer RPC's with invalid images references to HMI for processing.


joeljfischer:
@mrapitis I'm not clear on how this affects the Android / iOS libraries or RPC changes. Can you clarify if I'm missing something?

mrapitis:
The main impact would be between core and the HMI.  I'll update the PR.

PR #117, opened 2017-03-20
Use Boost library in SDL

AGaliuzov:
Writing proper failover code pledge to success. Automotive industry constantly focused on code quality and application performance. Usually development process consists of writing all kind of utilities, helper classes and local frameworks in order to have a good basis for the business logic implementation. A customer is usually cares about certain problem solution rather than whole bunch of helper utilities. But because of lack of ready solutions, developers had to reinvent the wheel over and over again. It certainly leads to increase of the number of errors, makes support harder and production code unstable. So the support becomes more expensive and solutions release rarely fit in time especially on early developments stages.

Unfortunately by default pre `C++11` do not have ready solutions for daily routine i.e file system, threads and memory management. There is a Qt library which has a lot of useful features out of the box but at the same time it is quite hard to adopt this library to the already existed project which uses iso c++ and has significant codebase. Another solution is the Boost library.

### DIFF
--- a/proposals/NNNN-use-Boost-library.md
+++ b/proposals/NNNN-use-Boost-library.md
@@ -1,0 +1,53 @@
+# Use Boost library in SDL
+
+* Proposal: [SDL-NNNN](NNNN-use-Boost-library.md)
+* Author: [Alexandr Galiuzov](https://github.com/AGaliuzov)
+* Status: **Awaiting review**
+* Impacted Platforms: [Core]
+
+## Introduction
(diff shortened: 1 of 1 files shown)


joeljfischer:
@AGaliuzov I'm fixing your PR. Please make sure that your PR title matches your evolution proposal title and that the text matches your evolution proposal introduction.

AGaliuzov:
@joeljfischer please consider this one as ready for review

joeljfischer:
@AGaliuzov Please double check your PRs before requesting review ready. This one has an empty bullet point.

AGaliuzov:
@joeljfischer thanks for note. I have fixed this and couple more.

joeljfischer:
@AGaliuzov Note that your PR title and proposal title do not match, I have changed the PR title to match the proposal title

PR #118, opened 2017-03-20
Implement logger abstraction

AGaliuzov:
The good abstraction for the logger allows to easily substitute the logger library from one to another in case of necessity. Different system has different logging mechanism and SDL has to be easily adopted to use any logger library as a background sink.


AGaliuzov:
@joeljfischer this one is ready for review

PR #119, opened 2017-03-20
External Policy Manager

JackLivio:
This proposal is for the addition of an example external policy manager to the sdl core repository. The external policy manager is an application/service that runs on the vehicle’s head unit and is responsible for encrypting/decrypting policy table data when a policy table update takes place.


JackLivio:
Example policy manager python script:

https://gist.github.com/JackLivio/1bee9cc7d13c1b84129d3507e32b9923

PR #121, opened 2017-03-21
Seats, HMI Settings and Audio controls updates

OlesiaV:
Using passenger's remote-control mobile application user is able to control specified modules of the vehicle via remote-control functionality turned on through HMI in vehicle settings. Currently only "CLIMATE" and "RADIO" modules are available for remote-control.


OlesiaV:
@joeljfischer please consider this proposal as ready for review

OlesiaV:
Closed as not actual

PR #135, opened 2017-03-29
Manual layout for `SDLLockScreenViewController`

joeljfischer:
This feature is to turn all lock screen related layouts and assets into code. This would simplify developers’ implementation of SDL; they would no longer have to import SDL lock screen assets into their project separately from the code.

### DIFF
--- a/proposals/nnnn-manual-lockscreen-layout-ios.md
+++ b/proposals/nnnn-manual-lockscreen-layout-ios.md
@@ -1,0 +1,240 @@
+# Manual layout for `SDLLockScreenViewController`
+
+* Proposal: [SDL-NNNN](nnnn-manual-lockscreen-layout-ios.md)
+* Author: [Joel Fischer](https://github.com/joeljfischer)
+* Status: **Awaiting review**
+* Impacted Platforms: iOS
+
+## Introduction
+This feature is to turn all lock screen related layouts and assets into code. This would simplify developers’ implementation of SDL; they would no longer have to import SDL lock screen assets into their project separately from the code.
+
+## Motivation
+Currently, SDL lock screen assets, such as layout storyboards and images, have to be packaged separately from the rest of the iOS library. Developers implementing the iOS library therefore have separate work to do in order to import the assets into their project, particularly when developers use Cocoapods to manage their SDL iOS dependency. Furthermore, currently, if we wish to update our assets in any way, they have to be re-imported by developers.
+
+This is what is currently written in the iOS README:
+
+```markdown
+If you used CocoaPods to install the SDL SDK, you must complete the following steps to add the default lock screen resources to your project:
+
+1. Select your application's build target, go to Build Phases, Copy Bundle Resources.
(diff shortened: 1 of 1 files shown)


Toyota-BSalahat:
Hi Joel,

Is there a reason that migrating to using (NSBundle *)bundleForClass:(Class)aClass is not being considered? This would only require setting up a bundle for the SDL cocoapods, and should solve the asset/storyboard issue.

joeljfischer:
@Toyota-BSalahat We do use `bundleForClass`, you can search for it in the project. That, unfortunately, doesn't solve the issue. And Cocoapods is set up to generate resources, but they still need to be manually imported.

Toyota-BSalahat:
@joeljfischer Very strange. I do see that in the SDLBundle class taking a closer look. I know I have implemented frameworks in the past that had embedded images/xibs/storyboards without having to do a manual import. The process as I understand it is laid out pretty well in this article, but I am not sure if it is helpful in this particular case or not.

https://www.tapglue.com/blog/ios-frameworks-storyboards-swift/

joeljfischer:
@Toyota-BSalahat We thought it should work the way it's set up, but that hasn't been our experience. If you are able to play with it and make it work as is, by all means. I think there's still advantages to moving to all code, as noted in the proposal, but that would be a big win for the current way as well.

Toyota-BSalahat:
@joeljfischer I dug a little bit deeper into this, and I believe I have a working solution for loading the sdl bundle. This seems to alleviate the issue when testing with my test proxy implementation.

```
+ (nullable NSBundle *)sdlBundle {
    NSBundle *sdlBundle = nil;
    
    NSBundle *localSdlBundle = [NSBundle bundleForClass:[SDLManager class]];
    
    NSURL *sdlBundleURL = [localSdlBundle URLForResource:@"SmartDeviceLink" withExtension:@"bundle"];
    sdlBundle = [NSBundle bundleWithURL:sdlBundleURL];
    
    if (sdlBundle == nil) {
        @throw [NSException exceptionWithName:NSInternalInconsistencyException reason:@"SDL WARNING: The 'SmartDeviceLink.bundle' resources bundle was not found. If you are using cocoapods, try dragging the bundle from the SmartDeviceLink-iOS pod 'products' directory into your resources build phase. If this does not work, please go to the SDL slack at slack.smartdevicelink.com and explain your issue. You may disable the lockscreen in configuration to prevent this failure, for now." userInfo:nil];
    }

    return sdlBundle;
}

```
Should I open up a pull request?

Thanks

joeljfischer:
@Toyota-BSalahat It seems you were partially correct. However, your solution would break the framework for anyone using Carthage or manually importing the framework. I'm implementing a revised solution based on yours that will work in both cases. Thanks for the hard work! 😄

Toyota-BSalahat:
@joeljfischer Awesome! Glad to help. I should have mentioned that was a solution aimed towards Cocoapods to aid in readability. Glad to hear it can be tweaked to work with the other two installation methods. I am not as familiar with Carthage myself.

PR #137, opened 2017-04-03
DTLS encryption

mrapitis:
This proposal is to add DTLS (Datagram Transport Layer Security) to SDL core.   SDL core will utilize the DTLS communications protocol as a preferred solution when sending encrypted data over protected services (such audio 0x0a  and video 0x0b ) including the handshake, encryption and decryption processes.


mrapitis:
@theresalech proposal is ready for review

PR #138, opened 2017-04-03
Gesture cancellation on video streaming

jhludwig:
This proposal defines a gesture cancellation event to prevent false gesture recognition during an interrupted video stream.

### DIFF
--- a/proposals/nnnn-touch-cancellation.md
+++ b/proposals/nnnn-touch-cancellation.md
@@ -1,0 +1,77 @@
+# Gesture cancellation on video streaming
+
+* Proposal: [SDL-NNNN](NNNN-filename.md)
+* Author: Masato Ogawa
+* Status: **Awaiting review**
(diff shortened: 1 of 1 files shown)


jhludwig:
@theresalech this is ready for review.

proposals/nnnn-touch-cancellation.md
```diff
@@ -0,0 +1,77 @@
...
+              ]
+  }
+}
+```
+
+## Potential downsides
+
+1) Can it be implemented with currently defined values?
+
+Yes, we can treat TouchType::BEGIN during TouchType::MOVE as a cancellation. However, to avoid misunderstanding, it good to define a single event rather than defining a sequence. And general mobile OS has an event like this.
```
theresalech: Please change "it" to "it's"

jhludwig: done, thx

PR #149, opened 2017-04-11
Add API Patch Version

jacobkeeler:
This proposal is for the addition of a patch version into the Mobile and HMI APIs, as well as the SyncMsgVersion struct.


jacobkeeler:
@theresalech ready for review

PR #156, opened 2017-04-18
Connectivity via iAP-BT and Transport Switch

robinmk:
This feature allows iOS devices to be connected with the Head Unit via Bluetooth without requiring a wired USB connection for registering media and non-media apps and also provides provision for transitioning to USB connection once the iOS device is also connected via USB.


robinmk:
@theresalech
Proposal is ready for review.

PR #163, opened 2017-04-25
Add zeroconf capability to WiFi transport

jhludwig:
The aim of adding zeroconf capability is to facilitate users configuring WiFi transport settings. Specifying head unit's IP address manually in car does not make sense in terms of user experience.


jhludwig:
@theresalech ready for review

PR #168, opened 2017-04-28
Template image 

acerati1:
This proposal is to enable the HMI to customize icons used by apps and make them match the graphical design of the HMI


acerati1:
@theresalech  the original proposal contains pictures which help understand the feature, please advise a location we can use to upload them.

kshala-ford:
Can we upload assets that are attached to a proposal to a subfolder? In this case we could upload to the following folder:

```
proposals/imagetype-template/image_1.png
proposals/imagetype-template/image_2.png
```

joeljfischer:
I would think it would be something like `assets/proposal/something/image-1.png`, that "something" I'm not sure about though. It would be nice to have it be the `SDL-0001` for easy correlation, but of course we don't know the number until it's pulled in. Maybe @theresalech or @joeygrover will have thoughts.

joeygrover:
I would suggest `assets/proposals/nnnn-imagetype-template/image-1.png` since we don't know the exact proposal number but we do know the proposal file name. 

Also this proposal has an incorrect filename and should be addressed.

kshala-ford:
@joeljfischer @joeygrover adding the number will require a change of the proposal file itself. I was hoping to avoid a second PR for every propsoal which includes images. Second I would rather go for `proposals/ assets/` than the other way around. It would reduce the relative path to the assets.

@joeygrover will change the file name to match the proposal name. Weren't aware they have to match after reading [README.md](https://github.com/smartdevicelink/sdl_evolution/blob/master/README.md) and [process.md](https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md)

joeljfischer:
@kshala-ford It wouldn't have to be a second PR, it would be the secretary's job (sorry @theresalech) to update those links to make them work. It would be highly annoying to have to update them, but it would make it much easier to find in the future.

I'd prefer to only keep proposals within the proposals folder, so that any service we built using Github's API could just iterate over the folder's contents instead of needing a special case. Reducing the relative path is nice, but not especially important, in my opinion.

That said, I'd rather not bikeshed over it too much.

The file name requirement is in the Pull Request Template that everyone has to read and delete before making a pull request:
> 1. Please make the title of your pull request the exact title of your proposal.

kshala-ford:
@joeljfischer I just wanted to propose an easy way of handling assets without potential source of error but I'm OK with the asset location if it enables you to do some automated actions in an easy way.

> That said, I'd rather not bikeshed over it too much.

Agree

> The file name requirement is in the Pull Request Template that everyone has to read and delete before making a pull request:

> > Please make the title of your pull request the exact title of your proposal.

But that's the proposal title which is the first line within your file. It's not clearly saying that the file name has to match. Seemed to be an unspoken rule.

joeljfischer:
> But that's the proposal title which is the first line within your file. It's not clearly saying that the file name has to match. Seemed to be an unspoken rule.

Ah gotcha, misread. I've generally kept them similar, but I don't think I've always kept them the same. But if that's what's going on moving forward, I can add it to the template? @joeygrover

acerati1:
> I would suggest assets/proposals/nnnn-imagetype-template/image-1.png since we don't know the exact proposal number but we do know the proposal file name.

done

> Also this proposal has an incorrect filename and should be addressed.

done
@theresalech ready for review

proposals/NNNN-template-images.md
```diff
@@ -0,0 +1,117 @@
...
+Example for all different modes:
+
+| Step | Day mode            | Night mode            | Highlighted mode            |
+|------|---------------------|-----------------------|-----------------------------|
+| 1.   | ![img][pattern-day] | ![img][pattern-night] | ![img][pattern-highlighted] |
+| 2.   | ![img][template]    | ![img][template]      | ![img][template]            |
+| 3.   | ![img][icon-day]    | ![img][icon-night]    | ![img][icon-highlighted]    |
+| 4.   | ![img][button-day]  | ![img][button-night]  | ![img][button-highlighted]  |
+
+Whenever the UI changes the HMI has to recreate images currently visible on the screen.
```
theresalech: Can you please add a comma after "changes"?

proposals/NNNN-template-images.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Potential downside
+
+N/A
+
+## Impact on existing code
+
+The impact on mobile side is minimal. App developer only have to set `.isTemplate` whenever appropriate. It's backwards compatible to head units which don't support this feature. Those head unit would just ignore this new parameter.
```
theresalech: Please change "have" to "has"

proposals/NNNN-template-images.md
```diff
@@ -0,0 +1,117 @@
...
+
+The impact on SDL core is minimal except the change of the HMI and mobile API. Unit tests may be requried but as this parameter is not used by core the effort should be trivial.
+
+The HMI has to add all the logic of template images. The image manipulation is surprisingly easy these day on commong UI frameworks.
```
theresalech: Please change last sentence to "...these days on common UI frameworks."

acerati1:
@theresalech changes applied

PR #173, opened 2017-05-08
Locale support

kshala-ford:
This proposal is about changing the way how SDL treats languages and localizations. The idea is to replace the `Language` enum and use the Locale structure/class of the native SDKs.


kshala-ford:
@theresalech The proposal is ready for review.

PR #174, opened 2017-05-08
Display name parameter

kshala-ford:
This proposal is about deprecating the parameter `displayType` and replace it with `displayName`.


kshala-ford:
@theresalech ready for review.

Issue #169, opened 2017-05-03
[Deferred] SDL 0057 - Add zeroconf capability to WiFi transport


theresalech:
Hello SDL community,

The review of "Add zeroconf capability to WiFi transport" begins now and runs through May 9, 2017. The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0057-add-zeroconf-capability.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/169

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Theresa Lech

Program Manager - Livio
theresa@livio.io

joeljfischer:
I have concerns about this proposal.
1. It takes as an underlying assumption that wi-fi should become a production use case, but I do not believe it should be. On iOS, and likely soon on android with recent changes, it is impossible to run in the background for any length of time, and it is impossible to start an app in the background without iAP. Therefore, wi-fi is not a production use case on iOS at a minimum, and likely not on Android in the near future.
2. This proposal is inadequate for the proposed functionality. It describes a single high-level mobile API to use zeroconf but this API is outdated and deprecated (since v4.3), does not describe if this is a replacement for the current IAP or if both are active, does not describe any of the lower level mobile changes necessary, only describes iOS, puts the burden on app developers to specify parameters they do not know and should not care about, does not speak of UI challenges when multiple found devices are present, requires the app to be in the foreground and does not describe the background challenges, and does not describe how this would be accomplished on Android or Core.

[...]

PR #181, opened 2017-05-15
Add zeroconf capability to WiFi transport

jhludwig:
The aim of adding zeroconf capability is to facilitate users configuring WiFi transport settings. Specifying head unit's IP address manually in car does not make sense in terms of user experience.

### DIFF
--- a/proposals/0057-add-zeroconf-capability.md
+++ b/proposals/0057-add-zeroconf-capability.md
@@ -1,9 +1,9 @@
 # Add zeroconf capability to WiFi transport
 
 * Proposal: [SDL-0057](0057-add-zeroconf-capability.md)
-* Author: Shinichi Watanabe
-* Status: **Deferred**
-* Impacted Platforms: [Core / iOS / Android / RPC]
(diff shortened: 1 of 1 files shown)


jhludwig:
@theresalech ready for review.  These changes are to respond to comments in the forum.

theresalech:
Hi @jhludwig - I was under the impression that the Steering Committee would be having a workshop to revise #169 together.  If it works for the Xevo team, I think we should bring these revisions to that workshop to discuss as a group.

jhludwig:
Yes we agree, we were simply making these changes available to that discussion.  it is fine to leave them unmerged if you like, we just wanted to share them.

shoamano83:
As I noted in https://github.com/smartdevicelink/sdl_evolution/issues/169, we have a newer proposal document for Wi-Fi Service Discovery (reference: https://github.com/smartdevicelink/sdl_evolution/issues/407). If we continue the discussion, it should be based on https://github.com/smartdevicelink/sdl_evolution/issues/407.
I think this PR is outdated, and can be closed now.

PR #193, opened 2017-06-02
Add H.264 over RTP format support for video streaming

shoamano83:
This proposal aims to achieve more robust and smooth video streaming for navigation and projection by adding timestamp information to a video stream.


shoamano83:
Hi @theresalech, I've revised the proposal to align with SDL-0058. It is ready for merge.

The change is to remove duplicate sections (video capabilities and format negotiation) from this proposal since they are now covered by SDL-0058. As a result, this proposal has impacts on iOS and Android proxies only. Please check the "Impacted Platforms" field.

Thanks.

PR #194, opened 2017-06-08
Idea To Implementation

joeygrover:
This proposal is to define a process on how ideas and improvements are taken from the SDL Evolution Process to actually being implemented, specifically for changes that alter the RPC spec.


joeygrover:
@theresalech please mark ready for review

PR #199, opened 2017-06-17
Notification of iAP transport type

shoamano83:
This proposal aims to notify underlying transport type to iOS SDL apps. The apps can use such notification to control, for example, the bitrate of video stream when running navigation.


shoamano83:
Hi @theresalech, this is ready for review. Thanks.

PR #203, opened 2017-06-23
New vehicle data - FuelRange

robinmk:
This proposal is about adding a new vehicle data to get the distance a vehicle can travel with the current level of fuel.


robinmk:
@theresalech, The proposal is ready for review.

PR #224, opened 2017-07-14
Support for multi session protocol string

robinmk:
This proposal introduces a new protocol string for head units which support iOS multisession (Ability for multiple apps to connect over the same protocol string) and also identifies the changes to the iOS proxy layer.


robinmk:
@theresalech , please mark as ready for review.

Issue #179, opened 2017-05-10
[Returned for Revisions] SDL 0061 - Locale support


theresalech:
Hello SDL community,

The review of the revised proposal "SDL 0061 - Locale support" begins now and runs through August 29, 2017.  The original review of "Locale support" occurred May 10 through May 16, 2017. The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0061-locale-support.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/179

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Theresa Lech

Program Manager - Livio
theresa@livio.io

joeljfischer:
I think this proposal needs to give the proposed changes to the mobile_api in XML form so that we can more clearly see these changes. It's not obvious given the proposal. I *think* what is being proposed is to deprecate the language enum (and turn them from required to optional) and all parameters that use it and to replace it with a string parameter which uses the locale, which the mobile libraries would turn into native objects.

Is the `locale` parameter optional, or required? It appears to be optional based on the text. If optional, we are left with two optional parameters, and it's possible for neither to be set, which is an error.

I don't think this proposal is adequate for approval as is, but I think the concept is a good one.

brettywhite:
I have two issues with the attached proposal that I would like to see expanded upon. 

1. There is no mention of if the locale support in iOS or Android are the same or if they are different. As the goal is to fallback onto that, it needs to be explicitly defined in the proposal (not just links to the developer documentation). This is more of an issue for Core I would imagine, as they would need to expect possibly different parameters depending on the library used. 

[...]

PR #225, opened 2017-07-19
SDL 0061 Locale Support

kshala-ford:
This proposal is about changing the way how SDL treats languages and localizations. The idea is to replace the `Language` enum and use the Locale structure/class of the native SDKs.

This PR updates the original proposal based on #179.


kshala-ford:
@theresalech the PR is ready for review

PR #227, opened 2017-07-19
Updated 0078 

joeygrover:
Revisions based on last night's steering committee call.

With the acceptance of [Constructed Payloads](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0052-constructed-payloads.md) for Control Frames we need to ensure we spec out all the changes that will be needed for the first version of these payloads. This includes adding a new param to the `StartService` frame for RPC services that contains the max version supported for control frame payloads.


joeygrover:
@theresalech ready for review

PR #233, opened 2017-07-31
SDLInterfaceManager

Toyota-BSalahat:
The SendHapticData RPC will allow video projection apps to tell SDLCore where their controls are located. This allows VPM apps to integrate with the OEM haptic engine in a way similar to template SDL apps. This proposal aims to implement the necessary supporting proxy logic so that applications can easily utilize the OEM haptic feedback mechanism.

### DIFF
--- a/proposals/NNNN-SDLInterfaceManager
+++ b/proposals/NNNN-SDLInterfaceManager
@@ -1,0 +1,208 @@
+# SDLInterfaceManager
+
+* Proposal: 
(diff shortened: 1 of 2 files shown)


Toyota-BSalahat:

joeygrover:
Please add the `.md`file extension to the proposal and add the proposal name in the file itself as [seen in the template](https://github.com/smartdevicelink/sdl_evolution/blob/master/0000-template.md).

Toyota-BSalahat:
@joeygrover got it. Should be updated now. Thanks.

PR #238, opened 2017-08-04
SubMenu icon

kshala-ford:
This proposal is about extending the RPC AddSubMenu to include a menu icon.

### DIFF
--- a/proposals/NNNN-submenu-icon.md
+++ b/proposals/NNNN-submenu-icon.md
@@ -1,0 +1,121 @@
+# SubMenu icon
+
+* Proposal: [SDL-NNNN](NNNN-submenu-icon.md)
+* Author: [Kujtim Shala](https://github.com/kshala-ford)
+* Status: **Awaiting review**
+* Impacted Platforms: [Core / iOS / Android / RPC]
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


kshala-ford:
@theresalech: The proposal is ready for review

proposals/NNNN-submenu-icon.md
```diff
@@ -0,0 +1,121 @@
...
+
+<function name="AddCommand" messagetype="request">
+  :
+  <param name="cmdIcon" type="Common.Image" mandatory="false">
+    <description>Deprecated</description>
+  </param>
+</function>
+```
+
+### Backward compatibility
```
theresalech: Does "Backward compatibility" only apply to the alternative solution, or would this be applied to the proposed solution?  If it applies to the proposed solution, please update the header to ## so that it's not in the "Alternatives considered" section of the proposal.  If it only applies to the alternative solution, this is fine as is.

kshala-ford: @theresalech the sub-section backward compatibility is related to the alternatve solution.

theresalech: @kshala-ford thanks for confirming!

PR #243, opened 2017-08-11
Expandable Design for Proprietary Data Exchange

robinmk:
The SystemRequest/OnSystemRequest RPCs provide a mechanism for various components within the head unit to exchange data with app(s) written specifically to communicate with those components. With the growing popularity of SDL as a means of data exchange between head units and the outside world, there is a need to provide service to more such components. Today, the list of such components is limited to the 'RequestType' enum. Expanding this list everytime would not be an ideal solution.
This document proposes a new mechanism for an expandable design in such scenarios by introducing a new string field called 'RequestSubType' to be used along with a new 'RequestType' enum called 'OEM_SPECIFIC'.


robinmk:
@theresalech , this proposal is ready for review.

PR #245, opened 2017-08-15
Progress Bar Seek Feature

dportis:
The Seek feature is a proposed new feature that is meant to enhance the usage of the progress bar for media apps with seek capability. Giving the user the ability to use the progress bar to seek will allow users to freely move throughout a song or podcast with ease.


dportis:
@theresalech , this proposal is ready for review.

PR #251, opened 2017-08-17
Sequentially Send Multiple RPCs - iOS

joeljfischer:
This proposes to add an API allowing the developer to sequentially send multiple RPCs and to be notified when all have completed sending.


joeljfischer:

PR #253, opened 2017-08-17
Change `HapticSpatialData` parameter name to `hapticSpatialData`

jacobkeeler:
# Change HapticSpatialData parameter name to hapticSpatialData
* Altered Proposal [SDL-0075](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0075-HID-Support-Plug-in.md)
* Author: [Jacob Keeler](https://github.com/jacobkeeler)
* Status: **Awaiting review**
* Impacted Platforms: [Core / iOS / Android / RPC]

## Introduction

With the acceptance of [SDL-0075](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0075-HID-Support-Plug-in.md) we need to change the name of the `HapticSpatialData` parameter to camel case.

## Motivation

Every parameter name in the spec uses camel case, `HapticSpatialData` needs to be changed to maintain consistency

## Proposed solution

Change the name of the `HapticSpatialData` parameter to `hapticSpatialData`

## Potential downsides

N/A

## Impact on existing code

* None, as this code has not yet been implemented or released. It would simply modify the name of this parameter in comparison to the original proposal.

## Alternatives considered

* None

PR #254, opened 2017-08-18
Change name of `SpatialStruct` to `Rect`

joeljfischer:
# Change HapticSpatialData parameter name to hapticSpatialData
* Altered Proposal [SDL-0075](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0075-HID-Support-Plug-in.md)
* Author: [Joel Fischer](https://github.com/joeljfischer)
* Status: **Awaiting review**
* Impacted Platforms: [Core / iOS / Android / RPC]

## Introduction

With the acceptance of [SDL-0075](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0075-HID-Support-Plug-in.md) we should change the name of the `SpatialStruct` struct to `Rect`.

## Motivation

We don't typically describe structs with `Struct` in the actual name, but `Spatial` doesn't work alone. Since this is simply a typical rect, with the addition of an optional `id`, we should keep it for general purpose rect usage.

## Proposed solution

Change the name of the `SpatialStruct` to `Rect`

## Potential downsides

N/A

## Impact on existing code

* None, as this code has not yet been implemented or released. It would simply modify the name of this parameter in comparison to the original proposal.

## Alternatives considered

* None

### DIFF
--- a/proposals/0075-HID-Support-Plug-in.md
+++ b/proposals/0075-HID-Support-Plug-in.md
@@ -69,7 +69,7 @@ This RPC would be the standardized SDL interface for haptic events, and would be
 
 ```xml
 
-<struct name="SpatialStruct">
+<struct name="Rect">
   <description>Defines spatial for each user control object for video streaming application</description>
   <param name="id" type="Integer" minvalue="0" maxvalue="2000000000" mandatory="false">
     <description>A user control spatial identifier</description>
@@ -132,7 +132,7 @@ This RPC would be the standardized SDL interface for haptic events, and would be
 <interface name="Common" version="x.x.x" date="yyyy-mm-dd">
 
 <struct name="Rect">
-  <param name="id" type="Integer" minvalue="0" maxvalue="2000000000" mandatory="true">
+  <param name="id" type="Integer" minvalue="0" maxvalue="2000000000" mandatory="false">
     <description>A user control's identifier.
     </description>
   </param>
--- a/proposals/0075-HID-Support-Plug-in.md
+++ b/proposals/0075-HID-Support-Plug-in.md
@@ -131,7 +131,7 @@ This RPC would be the standardized SDL interface for haptic events, and would be
 
 <interface name="Common" version="x.x.x" date="yyyy-mm-dd">
 
-<struct name="SpatialStruct">
+<struct name="Rect">
   <param name="id" type="Integer" minvalue="0" maxvalue="2000000000" mandatory="true">
     <description>A user control's identifier.
     </description>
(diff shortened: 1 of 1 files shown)


joeljfischer:
@theresalech ready for review

LuxoftAKutsan:
@joeljfischer SpatialStruct has `id` field. Should it be additional parameter in SendHapticData?

joeljfischer:
@LuxoftAKutsan That is true, however, it cannot be moved to `SendHapticData` because `SendHapticData` has multiple `SpatialStruct`s that each need a unique id. My current thought is to change the `id` parameter to optional, and uses of `Rect` that don't need an `id` will just not set that parameter.

LuxoftAKutsan:
@joeljfischer What about leaving such SpatialStruct  : 
```xml
<struct name="SpatialStruct">
  <description>Defines spatial for each user control object for video streaming application</description>
  <param name="id" type="Integer" minvalue="0" maxvalue="2000000000" mandatory="true">
    <description>A user control spatial identifier</description>
  </param>
  <param name="rect" type="Rect" mandatory="true">
    <description>The rectangle of the user control</description>
  </param>
</struct>
```

joeljfischer:
@LuxoftAKutsan I don't think so, because one of the primary problems is that `SpatialStruct` is not a name that fits with the project and isn't particularly descriptive. While it's not perfect to have an optional `id` on a rect, I don't think it's particularly bad either.

LuxoftAKutsan:
@joeljfischer not sure, that identifier in simple Rect is good. And if it will be optional it adds additional problems. For example we need identifiers of rect in SendHapticData requests, but identifiers are optional, and if some of them are missed, we won't have an error during request validation by API schema. And need do additional check. 
API is very hight level communication protocol, and changes in it rather expensive, so it shoud be designed as good as possible in my opinion. 
I agree that SpatialStruct is not the best name. May be we can try just to invent better name? 
If request named SendHapticData, we can name data as HapticData, can't we?

LuxoftAKutsan:
@joeljfischer also looks like this PR has wrong name ( same as https://github.com/smartdevicelink/sdl_evolution/pull/253)

theresalech:
The Steering Committee agreed to accept this proposal, which makes the following changes to the accepted SDL 0075 (OEM specific Human Interface Device support as Plug-in architecture in SDL proxy) proposal:
- Change `SpatialStruct` to `HapticRect`
- Change `SendHapticData` parameter name to `HapticRectData`
- Remove the xy width and height from `HapticRect` struct into its own struct called `Rectangle`
- Change descriptions to better reflect `Rectangle`

PR #267, opened 2017-08-25
SDLHapticHitTesterProtocol

Toyota-BSalahat:
SDL-0081 introduced the SDL Haptic Interface Manager. Since this manager stores the location of all haptic UI components, it can be used to ease integration with apps that utilize it when responding to touch events. This proposal aims to introduce a protocol that takes a SDLTouch object, and returns the selected haptic view when applicable.


Toyota-BSalahat:
@theresalech this is the proposal to add additional functionality to the haptic manager that we discussed in yesterday's workshop. It should be ready for review.

PR #272, opened 2017-08-31
Security ATF feature

LuxoftAKutsan:
This proposal describes support of security sessions in ATF (Automated Test Framework). ATF should be able to test SDL security feature.

Main ATF features for checking SDL security feature:

    Support of secure SDL sessions;
    Support of testing broken handshake;
    Support of sending raw data in secure channel;
    Support of expected secure requests\responses\notifications;
    Support Transport Layer Security (TLS) and Datagram Transport Layer Security (DTLS).


LuxoftAKutsan:
Not actual. ATF modification should not be implemented throw proposals

PR #276, opened 2017-09-05
SDL Policy Server Enhancements - UI Framework

crokita:
This proposal provides an analysis of possible front-end frameworks for the UI component of the SDL Policy Server, and suggests which one would make the most sense to use for this project.


crokita:
The proposal is ready for review @theresalech

PR #284, opened 2017-09-12
Tire pressure additions

kshala-ford:
The proposal is about extending the existing tire pressure API by providing the actual pressure value and the TPMS status of all mounted tires.


kshala-ford:
@theresalech The proposal is ready for review.

PR #292, opened 2017-09-19
New data transfer interface in Android Proxy for streaming

shoamano83:
The purpose of this proposal is to add new data transfer interface between Android Proxy and an app for video and audio streaming.


shoamano83:
Hi @theresalech, this proposal is ready for review. Thanks.

PR #294, opened 2017-09-19
Add Android SDL Internal Interface

joeygrover:
The internal SDL interface will help reduce the dependency on the SDLProxyBase class and allow us to move away from it in the future when we introduce a more enhanced framework.


joeygrover:
@theresalech ready for review

Issue #277, opened 2017-09-06
[Rejected] SDL 0091 SDLCarWindow Video Projection Developer Interface


theresalech:
Hello SDL community,

The review of the revised proposal "SDL 0091 - SDLCarWindow Video Projection Developer Interface" begins now and runs through October 17, 2017.  Reviews of previous iterations of "SDLCarWindow Video Projection Developer Interface" took place September 27 - October 3, 2017 and September 6 - September 12, 2017.  The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0091-SDLScreen-SDLWindow-Projection.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/277

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Theresa Lech

Program Manager - Livio
theresa@livio.io

joeljfischer:
> The app developer uses the carWindow property on the streaming media manager to obtain the SDLCarWindow instance. Then to begin capture, encoding and streaming of view hierarchy video frames, the developer assigns the root view controller to be projected to the streamingViewController property on the car window instance.

Would this be better as a change to the SDLStreamingMediaConfiguration? Or are they expected to be able to set the `streamingViewController` multiple times?

> SDLInterfaceManager

As I noted in my review for this component, I don't think this name is descriptive for what it does. A better name might be `SDLHapticItemLocator`.

> If the head unit communicates that it handles focus and selection, the SDLInterfaceManager will use the BOOL appHandlesTouches property on the SDLStreamingMediaConfiguration class to determine whether to iterate over the views contained in the SDLCarWindow streamingViewController property to determine and send the haptic spatial configuration to the head unit.

[...]

Issue #250, opened 2017-08-16
[Accepted] SDL 0084 - Progress Bar Seek Feature


theresalech:
Hello SDL community,

The review of "Progress Bar Seek Feature" begins now and runs through August 22, 2017. The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0084-Progress-Bar-Seek-Feature.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/250

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Theresa Lech

Program Manager - Livio
theresa@livio.io

stefanek22:
:+1:

joeljfischer:
I'm of a bit of two minds on this. The first is that I've definitely been annoyed in the car that I was unable to drag the slider to a later point in a podcast, and had to instead press the skip forward button many times to find the place I want. However, I can also see a case made that adding this increases driver distraction. Carplay I know does not have this feature, I don't know about Android Auto. On audiobooks or podcasts that are several hours long I can also envision a user getting frustrated at trying to get the seek bar to the *precise* point they want it.

A suggestion might be to have a specific number of points that the bar can skip to. For example, the bar by default allows 10 points along its length and the seek bar can only skip to those 10 points based on where the user lands his finger. This could be customizable to allow the app to define these points, for example, chapters in a podcast or audiobook.

The "infinite" number of point slider seems like a possible distracting element. I'm not going to give a thumbs up or down on this proposal as my reservations seem more like opinion to me than fact, and would be interested if there's any further information about the safety (or if other OEMs are using free seek bars) of this API.

mrapitis:
[...]

PR #305, opened 2017-10-03
0091-SDLScreen-SDLWindowProjection

GNXClone:
The iOS SDL framework implements an SDLCarWindow class to provide an easy-to-integrate developer interface to SDL video projection and the associated task of defining focusable and selectable UI elements.


GNXClone:
@theresalech The updated proposal is ready for review

theresalech:
@GNXClone SDL 0091 is currently in review: https://github.com/smartdevicelink/sdl_evolution/issues/277

Are you requesting that proposal be revised to this one?

GNXClone:
Yes. The revised 0091 replaces the old one. Let me address any outstanding questions and comments in #277.

proposals/0091-SDLScreen-SDLWindow-Projection.md
```diff
@@ -1,110 +1,209 @@
...
+The iOS SDL framework implements an SDLCarWindow class to provide an easy-to-integrate developer interface to SDL video projection the associated task of defining focusable and selectable UI elements and the translation of remote touch events into standard UIKit actions. 
 
 ## Motivation
 
-This proposal builds on the SDLStreamingMediaManager to provide a convenient and simple programming model for SDL app developers to project and enable interaction with the appropriate subset of their UIKit, OpenGL and hybrid UIKit/OpenGL user interfaces.
+Provide a convenient and simple programming model for SDL app developers to remotely project and enable interaction with the appropriate subset of their UIKit, OpenGL and hybrid UIKit/OpenGL user interfaces. [1]
```
theresalech: I'm not sure if the [1] is intended to link to the note at the end of the proposal, but it currently doesn't.

proposals/0091-SDLScreen-SDLWindow-Projection.md
```diff
@@ -1,110 +1,209 @@
...
-
-/**
-* Car Window instance responsible for providing video projection.
-* /
-@property (nonatomic, strong, readonly) SDLCarWindow *carWindow;
-:
-:
-```
-The SDLCarWindow interface is defined as follows:
+SDLCarWindow leverages SDLStreamingMediaManager and SDLTouchManager to hide the management of video projection and touch event handling from the app developer. For apps which use only UIKit derived UI elements, no SDL specific code is necessary. [1]
```
theresalech: Same comment regarding [1].

proposals/0091-SDLScreen-SDLWindow-Projection.md
```diff
@@ -1,110 +1,209 @@
...
 
 ## Introduction
 
-The iOS SDL proxy exposes the SDLCarWindow class to provide an easy-to-integrate developer interface to SDL video projection and the associated task of defining focusable and selectable UI elements. The SDLCarWindow attaches to the SDLStreamingMediaManager and exposes the SDLInterfaceManager described in the [SDLInterfaceManager](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0081-SDLInterfaceManager.md) proposal to allow the latter to determine the spatial arrangement of controls. The SDLInterfaceManager adopts the [SDLHapticHitTester](https://github.com/smartdevicelink/sdl_evolution/issues/270) protocol to enable correlation of SDLTouch events with the corresponding haptic region. In a later proposal, the SDLInterfaceManager will be extended to include focus and selection control on the handset, following the UIFocusEngine model from tvOS.
+The iOS SDL framework implements an SDLCarWindow class to provide an easy-to-integrate developer interface to SDL video projection the associated task of defining focusable and selectable UI elements and the translation of remote touch events into standard UIKit actions.
```
theresalech: I'm a little unclear on how this sentence is supposed to read.  Should there be a comma after "projection"?

proposals/0091-SDLScreen-SDLWindow-Projection.md
```diff
@@ -1,110 +1,209 @@
...
- *  View Controller that will be streamed.
- */
-@property (strong, nonatomic, nullable) UIViewController *streamingViewController;
+The video projection interface is exposed to the application as a UIScreen object representing a virtual external display. The external view is separate from the view rendered on the device's LCD. This allows the developer to optionally forego the lock screen, instead displaying a different interface on the device than what is projected onto the head unit.
 
-/**
- * Dimensions of the target in-car screen
- */
-@property (assign, nonatomic, readonly) CGSize screenDimensions;
+![Separate Displays](https://i.imgur.com/dIBQnyZ.png "Separate Displays")
```
theresalech: Could you please upload this image to the /assets/proposals folder and link to that image?  You can reference [SDL 0084](https://github.com/smartdevicelink/sdl_evolution/issues/250) as an example.

proposals/0091-SDLScreen-SDLWindow-Projection.md
```diff
@@ -1,110 +1,209 @@
...
 
-For head units that allow the app to manage focus and selection, the SDLInterfaceManager provides the local logic to do so and interacts with the views through the [UIFocusEnvironment protocol](https://developer.apple.com/documentation/uikit/uifocusenvironment).
- 
-The SDLCarWindow class will also include the previously app-level common logic that interacts with the existing SDLStreamingMediaManager class to stream pixel buffers and receive audio data. It relies on the helper class SDLStreamingMediaLifecycleManager to ensure streaming is allowed and manage the streaming state.
+Currently supported types are: UIButton, UIControl, UINavigationViewController, UITableView, UICollectionViewCell, UIScrollView, UISearchBar, UITextField, UISegmentedControl and UITabBarController. [2]
```
theresalech: [2] does not link to the note at the end of the proposal (not sure if this is intentional).

proposals/0091-SDLScreen-SDLWindow-Projection.md
```diff
@@ -1,110 +1,209 @@
...
 
-```objc
-extern NSString *const SDLVideoStreamDidStartNotification;
-extern NSString *const SDLVideoStreamDidStopNotification;
+Touch events for custom views defined by the app must be handled by the app. An app wishing to handle touch events must adopt and implement the SDLTouchManagerDelegate protocol. The app will then have the choice of handling a touch event, or passing the event on to SDL to handle it automatically. When the app handles the event, it returns YES from the delegate method. A return value of NO instructs SDL to handle the event.
 
-extern NSString *const SDLAudioStreamDidStartNotification;
-extern NSString *const SDLAudioStreamDidStopNotification;
+In a later proposal, SDL will be extended to include focus and selection control on the handset, following the UIFocusEngine model from tvOS. [3]
```
theresalech: [3] does not link to the note at the end of the proposal (not sure if this is intentional).

proposals/0091-SDLScreen-SDLWindow-Projection.md
```diff
@@ -1,110 +1,209 @@
...
 
 ## Impact on existing code
-
-This proposal has no impact on existing video/audio streaming applications. It provides a new higher level interface that should improve developer productivity and a foundation for both touch and physical button focus/select interactions.
+* Apps adopting SDLTouchManagerDelegate will need to return BOOL instead of void for delegate methods.
+* SDLTouchManager moves from SDLStreamingMediaLifecycleManager to SDLCarWindow
+* SDLCarWindow deprecates SDLHapticManager. Functionality previously found in SDLHapticManager has been moved into SDLCarWindow.  
 
 ## Alternatives considered
+* Force the each application to handle touch events via SDLTouchManagerDelegate. The developer has to handle the translation of a view and location into the desired action. Touch management will be inconsist amongst SDL apps, which could lead to user dismay.
```
theresalech: I believe "inconsist" should be updated to "inconsistent".

proposals/0091-SDLScreen-SDLWindow-Projection.md
```diff
@@ -1,110 +1,209 @@
 # SDLCarWindow Video Projection Developer Interface
 
 * Proposal: [SDL-0091](0091-SDLScreen-SDLWindow-Projection.md)
-* Author: [David Switzer](https://github.com/davidswi)
-* Status: **Returned for Revisions**
+* Author: [Michael Pitts](mailto:mpitts@xevo.com)
```
theresalech: Can you please update this to your github account?

proposals/0091-SDLScreen-SDLWindow-Projection.md
GNXClone: Inadvertently deleted the word "and" from the original text. It's been added back into the text.

proposals/0091-SDLScreen-SDLWindow-Projection.md
GNXClone: Still new to this markup syntax. I don't know how to do that yet. An example would be great.

proposals/0091-SDLScreen-SDLWindow-Projection.md
GNXClone: We've decided it would be best to move touch manager related changes to another proposal. This entire paragraph goes away.

proposals/0091-SDLScreen-SDLWindow-Projection.md
GNXClone: Done

proposals/0091-SDLScreen-SDLWindow-Projection.md
GNXClone: The revised proposal does away with "notes"

proposals/0091-SDLScreen-SDLWindow-Projection.md
GNXClone: Done. Will I need a separate PR for this graphic?

proposals/0091-SDLScreen-SDLWindow-Projection.md
GNXClone: Removed notes

proposals/0091-SDLScreen-SDLWindow-Projection.md
GNXClone: Removed notes

GNXClone:
I think all comments have been addressed

joeljfischer:
I've been reviewing the new proposal:

> [UIScreen screens] is overidden to add the SDL managed UIScreen to the list of available screens.

Does this mean overriding or swizzling? Because overriding `UIScreen` wouldn't work this way, right? (It would have to be something like `SDLScreen`). The reason I bring this up is because the proposal states that no SDL specific code is needed if they use the car window API. I don't think that will be the case due to the need to override things like `UIScreen` and `UIWindow`.

I'm also not sure what's happening in the `updateCarWindow` method code. I don't know what type of object the `sdlWindow` property is supposed to be. I'm not sure what the purpose of the storyboard is. I'm not sure how it's pulling the screen from `[UIScreen screens]`. Perhaps by "overriding" you mean "swizzling"?

> Ownership of SDLTouchManager moves from SDLStreamingMediaLifecycleManager to SDLCarWindow. SDLStreamingMediaLifecycleManager links to SDLCarWindow to retain access to SDLTouchManager.

I think that the `SDLTouchManager` should stay on That would move it for everyone, whether they use Car Window or not. Instead, it should be dependency injected into car window when it's created and held weakly.

> At the time of this proposal, the app handles touch events by adopting the SDLTouchManagerDelegate protocol. In a later proposal, SDLTouchManager will be extended to automatically handle touch events for any UIResponder derived class, removing this burden from the app developer and deprecating SDLTouchManagerDelegate.

Is this still intended for 5.1 or (as I would recommend) post-5.1?

> SDLCarWindow walks the view hierarchy to build a list of focusable views. This list is sent to the head unit as SDLHapticRectData.

So CarWindow is intended to replace haptic manager? Would this deprecate the haptic manager and everything that's done in it?

> SDLCarWindow automatically handles view layout updates by listening for calls to -[CALayer layoutSublayers].

> SDLCarWindow deprecates SDLHapticManager. Functionality previously found in SDLHapticManager has been moved into SDLCarWindow.

Why is this re-doing everything already done in the haptic manager, except with swizzling? The current one seems pretty good without it, what are the pros/cons of this compared to the current implementation. I would think that the haptic manager is a good separation of responsibilities and would continue to exist, though perhaps in the `SDLCarWindow` class (as a property). This deprecation is going to really suck though as to get it into 5.0 it had to be integrated pretty strongly into the `SDLStreamingLifecycleManager`. The functionality has to remain. The deprecation process to shift from 5.0 to CarWindow in 5.1 is going to be a hairpull, which is exactly what I was trying to avoid in previous workshops. It's going to be difficult to shift from the 5.0 methods of doing things to CarWindow while retaining 5.0's functionality.

> In a later proposal, SDL will be extended to include focus and selection control on the handset, following the UIFocusEngine model from tvOS.

Minor point, but from a wording perspective, this hasn't been agreed to, so it should probably say something more like: "Focus and selection control on the handset, following the UIFocusEngine model from tvOS will be proposed at a later date and is not contained in this proposal."

> The projected video view is separate from the view on the device's LCD allowing for optional use of device while projecting.

But only when driver distraction allows. Most of the time it will be a lock screen. Performance would likely suffer if we did it this way.

> SDLCarWindow uses swizzling. However, all swizzled methods are public and have not changed since iOS 2.0-3.2.

Additionally, because we are a library swizzling Apple methods, what are the ramifications of swizzling if the dev is already swizzling? Are there going to be devs that balk at using a lib that swizzles?

*Additional Notes:*
* This doesn't actually give us a proposed interface for `SDLCarWindow`, just a collection of implementation details and notes on how it will be implemented. Which are important, but we need an actual proposed interface and changes to other public interfaces required to make this proposal happen.

Toyota-BSalahat:
I share @joeljfischer's concern about how the UIScreen is being managed here, the proposal doesn't really detail what "override" means.

```objc
// When an external screen becomes available, attach a storyboard  
// designed specifically for the head unit's display.
- (void) updateCarWindow:(NSNotification *)notification
{
    if (UIScreen.screens.count <= 1) {
        self.sdlWindow = nil;
        return;
    }

    UIScreen*           sdlScreen   = [UIScreen screens][1]; //1
    UIStoryboard*       storyboard  = [UIStoryboard storyboardWithName:@"SDLWindow" bundle:[NSBundle mainBundle]];//2
    UIViewController*   vc          = [storyboard instantiateInitialViewController]; 

    vc.view.frame  = sdlScreen.bounds;
    vc.view.bounds = sdlScreen.bounds;

    self.sdlWindow = [[UIWindow alloc] initWithFrame:sdlScreen.bounds];
    self.sdlWindow.screen = sdlScreen;
    self.sdlWindow.rootViewController = vc;
    [self.sdlWindow addSubview:vc.view];
    [self.sdlWindow makeKeyAndVisible];
}
```

I have some concerns where I have commented 1 and 2 above.

1. Why not have a SDL subclass of UIScreen that we can look for using a filter method, instead of grabbing a hard index? What if SDL's screen is not present at index [1]?

2. Are we expecting that each app will have a SDLWindow storyboard? Why not allow the app to inject a view controller as an argument instead?

> Focusable and selectable UI elements

SDLCarWindow walks the view hierarchy to build a list of focusable views. This list is sent to the head unit as SDLHapticRectData.

The interface manager already contains this functionality. Is there a benefit trying to put it in SDLCarWindow? It seems like that would break encapsulation. If there needs to be changes to the crawling logic, I think those changes should be proposed inside of the interface manager class/protocol, not the car window.

> SDLCarWindow automatically handles view layout updates by listening for calls to -[CALayer layoutSublayers].

Is this implying use of Swizzling?

PR #317, opened 2017-11-03
AutoCompleteList

dportis:
The AutoCompleteList will give apps the ability to have multiple listing options while searching for things like destination and information; instead of giving a single suggestion.


dportis:
@theresalech Please mark as "review ready"

PR #318, opened 2017-11-06
SetAudioSreamingIndicator RPC

kshala-ford:
The proposal is about adding a new RPC for media apps to indicate the action of a play/pause button.


kshala-ford:
@theresalech PR is ready for review.

PR #329, opened 2017-11-08
Policy validation of RPCs referenced in multiple functional groupings

dboltovskyi:
Current SDL implementation of PRCs policy check includes the following simple flow. Application could have several functional groupings assigned to it and RPC could be a member of several groupings. Moreover, in each grouping RPC could have different sets of parameters allowed to be processed. In case of user disallows some grouping, it will add restrictions for processing RPC with allowed parameters if request contains also some disallowed.

A new idea is to implement extended policy check for the case when some RPC is included into different functional groupings and such groupings have different user permissions.


AByzhynar:
@robinmk Proposal updated. SDL behavior generalized for all RPCs.

theresalech:
The Steering Committee voted to close this pull request due to inactivity.  The author may re-open the PR when the proposal is ready to be revisited.

PR #339, opened 2017-11-13
SDL iOS Move RPC Notifications to a Concurrent Background Queue

joeljfischer:
This proposal seeks to move iOS RPC notifications from being sent on the main thread to a concurrent background queue.


joeljfischer:
@theresalech ready for review

NickFromAmazon:
Will this work for notifications for which order is important? `OnKeyboardInput` and `OnAudioPassThru` come to mind.  If the main goal is not to block the main thread, would it make more sense to use a background serial queue?

Glancing at the code, it also appears that the current implementation of the SDLManager may be not threadsafe (I see requests are stored in a dictionary in SDLResponseDispatcher without any obvious synchronization?)  I'd assume this would be a pre-req for this change, a pretty typical use case would be to make an additional SDL request in response to these notifications.

Apologies I'm misreading the code, or this is already covered under another item that I'm missing.

Issue #269, opened 2017-08-30
[Accepted with Revisions] SDL 0089 - Mobile API versioning


theresalech:
Hello SDL community,

The review of the revised proposal "SDL 0089 - Mobile API versioning" begins now and runs through September 19, 2017.  The original review of "Mobile API versioning" occurred August 29 through September 5, 2017.  The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0089-mobile-api-versioning.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/269

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Theresa Lech

Program Manager - Livio
theresa@livio.io

joeljfischer:
I have a few notes:
1. I'm not sure that the `minVersion` attribute on the spec as a whole is necessary as the spec is under semantic versioning; therefore, the spec *should* always be backward compatible to the last major version change. This hasn't really been the case until v4.0, but it is at this point.

2. Do params that are introduced with an RPC need the `added` tag or if it's missing can it be assumed to have been added when the RPC is added? It's not a big deal to add the version to everything, just wondering if anyone thinks it might be cleaner. (I just now noticed this in the alternatives as well)

3. You note in the section "Parameter element signature" that _any_ change would be to cause a new parameter to be added with an "added / removed" pair. I think this should only occur in the case of a breaking change, but not in a minor / widening change. You give the example:

```xml
<struct name="Choice">
  <param name="vrCommands" type="String" minsize="1" maxsize="100" maxlength="99" array="true" removed="4.5.0" /> <!-- implicitly mandatory -->
[...]

PR #340, opened 2017-11-14
Mobile API versioning revision

kshala-ford:
The proposal to add versioning to the mobile API was accepted with the revision to use the maintainer’s proposal. The suggested code will solve the same problem, while making parsing easier. The suggested code is 

```xml
<enum name="ButtonName">
</element>
  <element name="OK" mandatory="false" deleted="4.6.0">
    <description>
      The button name for the physical OK button.
      It is not related to the Play/Pause button.
      Please use the physical `PLAY_PAUSE` button.
    </description>
    <previousVersions>
      <element name="OK" mandatory="true" since="1.0.0" removed="4.5.0"/>
      <element name="OK" mandatory="false" since="4.5.0" deprecated="4.5.2"/>
      <element name="OK" mandatory="false" deprecated="4.5.2" removed="4.6.0"/>
    </previousVersions>     
  </element>
</enum>
```

With the following comment

> I think this is much more confusing and difficult to write a parser for than the maintainer suggestion above. The maintainer suggestion provides only two version parameters: from and until (though these could be different, such as since). Then, all the other parameters become booleans, including deprecated and removed or deleted (whichever wording is chosen). To me, this seems easy to read and easy to parse. In particular, the mandatory change in the revised version seems very confusing.

This document will show scenarios for the mobile API and how to solve them using the author’s proposal and the maintainer’s proposal. At the end both proposals will be compared:

-	What’s the increase the mobile API?
-	What’s the complexity to read the mobile API as a human?
-	What’s the complexity to read the mobile API as a parser?

To compare the parser complexity the comparison includes xpaths to locate the relevant information including versioning.


kshala-ford:
@theresalech PR ready for review.

LuxoftAKutsan:
@kshala-ford this proposal is about integrating into MOBILE API structure version control system for each element. 
Not sure that is would be rely useful. As for me changing in API's is major changes and should be listed in release notes.

kshala-ford:
@LuxoftAKutsan actually this proposal is already accepted (see #269) and allows to store versioning metadata in a structured way. Once this information is available it's planned to use that metadata for various reasons e.g. to avoid version clashes between app and head unit. This will be done in future proposals.

jacobkeeler:
@kshala-ford Sorry if I'm misunderstanding, but as far as I can tell, the proposed solution by the maintainer is:

```xml
<enum name="ButtonName">
  <element name="OK" mandatory="false" deleted="true" from="4.6.0">
    <description>
      The button name for the physical OK button.
      It is not related to the Play/Pause button.
      Please use the physical `PLAY_PAUSE` button.
    </description>
    <previousVersions>
      <element name="OK" mandatory="true" from="1.0.0" until="4.5.0"/>
      <element name="OK" mandatory="false" from="4.5.0" until="4.5.2"/>
      <element name="OK" mandatory="false" deprecated="true" from="4.5.2" until="4.6.0"/>
    </previousVersions>     
  </element>
</enum>
```

Rather than:
```xml
<enum name="ButtonName">
  <element name="OK" mandatory="false" deleted="4.6.0">
    <description>
      The button name for the physical OK button.
      It is not related to the Play/Pause button.
      Please use the physical `PLAY_PAUSE` button.
    </description>
    <previousVersions>
      <element name="OK" mandatory="true" since="1.0.0" removed="4.5.0"/>
      <element name="OK" mandatory="false" since="4.5.0" deprecated="4.5.2"/>
      <element name="OK" mandatory="false" deprecated="4.5.2" removed="4.6.0"/>
    </previousVersions>     
  </element>
</enum>
```

The latter was just a potential solution proposed by Joel, which was later revised. The revised version of this format was the accepted format.

The most recent version of an element is in it's base tag, and the entire state of that element is copied whenever a revision is made (resulting in a new entry in `previousVersions`). A range of versions (`from=... until=...`) is also included to show when that version of the tag was used. In this case, the state of the tag is kept with each revision, rather than just the changes since the previous version (as is the case with your proposed solution).

This solution makes it very easy to target old versions of the spec, and is very easy to parse. The parser can just use the copy of the element/parameter that is tied to the targeted RPC Spec version. It is certainly on the verbose side, however.

Based on the format which was accepted, the maintainer's proposal for Scenario 1 would be:
```xml
 <struct name="SyncMsgVersion">
   <description>Specifies the version number of the SYNC V4 protocol, that is supported by the mobile application</description>
   <param name="majorVersion" type="Integer" minvalue="1" maxvalue="10" />
   <param name="minorVersion" type="Integer" minvalue="0" maxvalue="1000" />
   <param name="patchVersion" type="Integer" minvalue="0" maxvalue="1000" mandatory="false" deleted="true" from="5.0.0">
     <previousVersions>
       <param name="patchVersion" type="Integer" minvalue="0" maxvalue="1000" mandatory="false" from="4.4.0" until="4.5.0">
       <param name="patchVersion" type="Integer" minvalue="0" maxvalue="1000" mandatory="false" deprecated="true"  from="4.5.0" until="5.0.0" />
     </previousVersions>
   </param>
 </struct>
```

Scenario 2:
```xml
 <struct name="SyncMsgVersion" deleted="true" from="6.0.0">
   <description>Specifies the version number of the SYNC V4 protocol, that is supported by the mobile application</description>
   <param name="majorVersion" type="Integer" minvalue="1" maxvalue="10" />
   <param name="minorVersion" type="Integer" minvalue="0" maxvalue="1000" />
   <param name="patchVersion" type="Integer" minvalue="0" maxvalue="1000" mandatory="false" deleted="true" from="5.0.0">
     <previousVersions>
       <param name="patchVersion" type="Integer" minvalue="0" maxvalue="1000" mandatory="false" from="4.4.0" until="4.5.0">
       <param name="patchVersion" type="Integer" minvalue="0" maxvalue="1000" mandatory="false" deprecated="true" from="4.5.0" until="5.0.0" />
     </previousVersions>
   </param>
   <previousVersions>
     <struct name="SyncMsgVersion" from="1.0.0" until="5.0.0" />
     <struct name="SyncMsgVersion" deprecated="true" from="5.0.0" until="6.0.0" />
   </previousVersions>
 </struct>
 
 <struct name="MsgVersion" from="5.0.0">
   <param name="majorVersion" type="Integer" minvalue="1" maxvalue="10" />
   <param name="minorVersion" type="Integer" minvalue="0" maxvalue="1000" />
 </struct>
```

Scenario 3:
```xml
<struct name="MsgVersion" from="5.0.0">
  <param name="majorVersion" type="Integer" minvalue="1" maxvalue="100" from="5.1.0">
    <previousVersions>
      <param name="majorVersion" type="Integer" minvalue="1" maxvalue="10" from="5.0.0" until="5.1.0" />
    </previousVersions>
  <param name="minorVersion" type="Integer" minvalue="0" maxvalue="1000" />
</struct>
```

Scenario 4:
```xml
<struct name="MsgVersion" from="5.0.0">
  <param name="majorVersion" type="Integer" minvalue="1" maxvalue="100" deleted="true" from="6.0.0">
    <previousVersions>
      <param name="majorVersion" type="Integer" minvalue="1" maxvalue="10" from="5.0.0" until="5.1.0" />
      <param name="majorVersion" type="Integer" minvalue="1" maxvalue="100" from="5.1.0" until="5.2.0" />
      <param name="majorVersion" type="Integer" minvalue="1" maxvalue="100" deprecated="true"  from="5.2.0" until="6.0.0" />
    </previousVersions>
  </param>
  <param name="minorVersion" type="Integer" minvalue="0" maxvalue="1000" />
</struct>
```

joeljfischer:
@jacobkeeler Thanks for listing this out! It appears correct to me, what you posted is what was accepted rather than the listing within this PR. The only scenario I can see being confusing for human readership is scenario 2 (though once you understand the structure, I think it's not too bad), but considering that the parsing is very easy due to just finding the version it is within in the `from` and `until` tags, while the rest are simple booleans. The only possible change I could see being useful is to, instead of `previousVersions` being from oldest to newest, making it newest to oldest, because the current version is the newest. I believe that the original proposal is still best.

kshala-ford:
I've modified the file based on the comment above. The maintainer's proposal should now match.

Different to your opinion I honestly don't see any advantage to the original proposal. The footprint is so much higher and it requires much more changes for the parser. The authors proposal only adds a minimum footprint by using a single attribute. It cannot be made more easy for humans and parsers to read those changes.

kshala-ford:
Hi @theresalech @joeljfischer @joeygrover , jut wanted to let you know that I updated the PR. I deleted my revision.md file and updated the original file to match what we agreed on. The versioning proposal should be ready now.

PR #342, opened 2017-11-14
Open Menu RPC

kshala-ford:
This proposal is about providing the capability to open the built-in app-menu which is provided by the native UI.


kshala-ford:
@theresalech Proposal ready for review.

PR #344, opened 2017-11-14
Default SoftButtonCapabilities

mrapitis:
This proposal is to update the iOS and Android mobile libraries with a specified default list size value for SoftButtonCapabilities when a headunit erroneously returns a SoftButtonCapabilities list that contains just one item. The purpose of the SoftButtonCapabilities list is to inform the mobile application how many SoftButtons are available for the utilized template and the capabilites of each SoftButton. The SoftButtonCapabilities list is received by the mobile application from the RegisterAppInterface and SetDisplayLayout RPC responses.


mrapitis:
@theresalech Proposal ready for review.

PR #349, opened 2017-11-17
Remote Control - OnRCStatus notification - Update according to review comments

yang1070:
One simple resource allocation rule is that at any given time only one application controls a RC module. For example, if SDL allocates a climate control module to application ONE, SDL rejects other applications to change the settings of the same climate control module. The module is freed when application ONE exits, disconnects, goes to HMI level NONE or no longer has the right to control the module due to various reasons (for example, a policy update revokes the access right to the module from the app, or SDL re-allocates the module to another app). In the last case, the mobile application is not aware of the change.

A new notification originated from SDL is nice to have. If SDL allocates a module to an application or it de-allocates a module from an application, SDL shall send OnRCStatus notifications to both the mobile application and the HMI. The notification shall contain an array of the RC modules allocated to the application, so that the mobile application can know it gains or loses control of a module before a failure request, and the HMI can update the status on which application controls which modules.


yang1070:
@theresalech This PR has the update to proposal SDL-0106 according to the review comments. It is ready for review. thank you.

theresalech:
Thanks @yang1070 - When we review SDL-0106 during our 2017-11-21 Steering Committee meeting, we will vote on accepting the proposal with these revisions.

yang1070:
@theresalech Got it. Thanks.

PR #353, opened 2017-11-27
Video Streaming Backgrounded String

joeljfischer:
Display a string of text on the head unit screen when the app leaves the foreground.


joeljfischer:
@theresalech Ready for review

PR #361, opened 2017-12-07
SDLImageUploadManager.md

Toyota-BSalahat:
Currently the logic SDL applications must implement to manage image uploads is tedious and error prone. The typical implementation is rewriting images using the same filename any time the image needs to be shown, or uploads all images at launch and never verifies their existence before calling a Show request. This proposal aims to centralize image management logic, optimize uploads with a local cache when possible, and generally move all image management responsbilities into the proxy.


Toyota-BSalahat:
@theresalech this should be ready for review

Toyota-BSalahat:
@theresalech I have updated the title.

PR #367, opened 2017-12-19
Add Secondary AppHMIType Array iOS

joeljfischer:
This proposal adds a new optional parameter to `SDLLifecycleConfiguration` allowing for more than one AppHMITypes when using `SDLManager`.


joeljfischer:
@theresalech Proposal is ready for review

PR #391, opened 2018-02-06
Android Manifest Metadata

joeygrover:
The current Android library relies on a lot of internal methods and functions that try and prevent developers from making changes to their apps. However, as Android becomes more and more strict on what system level methods are available the SDL library will need to move to more conventional methods to accomplish certain goals. One of the ways to achieve this is requiring developers to add metadata fields to the `SdlBroadcastReceiver`. This will allow the library to query all SDL enabled apps and retrieve attributes about them without having to perform any inter-app communication.


joeygrover:
@theresalech ready for review

PR #409, opened 2018-02-20
Driver Distraction Notification Upon Registration

joeljfischer:
This document proposes that Core send an additional Driver Distraction Notification just after registration to ensure that the mobile device knows the distraction status at all times.


joeljfischer:
@theresalech ready for review

PR #419, opened 2018-03-06
Template Improvements: Color Scheme

joeljfischer:
This proposal enables apps to provide a custom color scheme. This will allow OEMs to still strictly control layout and design while allowing apps some amount of branding and customization.


joeljfischer:
@theresalech ready for review

Issue #334, opened 2017-11-08
[Accepted with Revisinos] SDL 0109 - SetAudioStreamingIndicator RPC


theresalech:
Hello SDL community,

The review of "SetAudioStreamingIndicator RPC" begins now and runs through November 14, 2017.  The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0109-set-audio-streaming-indicator.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/334

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Theresa Lech

Program Manager - Livio
theresa@livio.io

JackLivio:
If the purpose of this proposal is to ensure that the play / pause button is being displayed properly, then this rpc is redundant since the HMI can use the setMediaClockTimer RPC to properly display the correct status of play or pause. 

If it is desired to have more options for the possible state of playback (stopped, buffering, etc), then I would suggest adding the new enum to the onHMIStatus notification instead of creating a whole new RPC. Also onHMIStatus already has a parameter related to audio streaming (audioStreamingState) so defining playback level within this RPC would make sense and would avoid the need to define a new RPC.

joeljfischer:
> If it is desired to have more options for the possible state of playback (stopped, buffering, etc), then I would suggest adding the new enum to the onHMIStatus notification instead of creating a whole new RPC. Also onHMIStatus already has a parameter related to audio streaming (audioStreamingState) so defining playback level within this RPC would make sense and would avoid the need to define a new RPC.

The `onHMIStatus` and `audioStreamingState` go from the head unit to the phone, whereas this RPC needs to go from phone to head unit.

joeljfischer:
[...]

PR #424, opened 2018-03-13
Update 0109 Set audio streaming indicator

kshala-ford:
Revision to SDL0109 #334


kshala-ford:
@theresalech see above. It's the revision for the accepted proposal. Ready for review.

PR #437, opened 2018-03-27
Support for Short and Full UUID App ID

mrapitis:
Some OEM legacy head units can't use an app id that has been generated by smartdevicelink.com. We would like to propose a shortened version of the UUID to be used in the legacy app id field and we add another variable for "fullAppID"


mrapitis:
@theresalech ready for review

PR #439, opened 2018-03-27
Add SDL Core Daemon script

jacobkeeler:
This proposal is for the addition of a daemon script allowing simple control over the lifecycle of SDL Core. This script would allow users to easily start/restart/stop SDL Core in the background without needing to create their own scripts to manage this behavior, as is currently the case.


jacobkeeler:
@theresalech Ready for review

PR #466, opened 2018-04-18
Define Handling of Duplicate Correlation IDs

jacobkeeler:
This proposal is for the clarification of the SDL protocol behavior in the case that an app sends multiple messages with the same correlation ID.


jacobkeeler:
@theresalech Ready for review

PR #467, opened 2018-04-19
Make spaceAvailable field non-mandatory

jacobkeeler:
This proposal is to make the `spaceAvailable` field in the `PutFile`, `DeleteFile`, and `ListFiles` responses non-mandatory in the RPC spec.


jacobkeeler:
@theresalech Ready for review

PR #472, opened 2018-04-23
Update 0158-cloud-app-transport-adapter.md with revisions

JackLivio:
Revisions include usage of OEM App store for app selection and authentication, retry connection attempts, opening websocket connections only when an app is activated, usage of a hashed VIN.


proposals/0158-cloud-app-transport-adapter.md
```diff
@@ -71,35 +68,133 @@ This is what the app_policies section of a policy table that includes a web app
...
 ```
 
-The endpoint field includes the URL/IP and port that the SDL web app can be connected to. The certificate field is used for secured RPC service connections which equates to opening a secured websocket connection. The TLS "certificate" field could be optional to allow for unsecured RPC service connections.
+The `endpoint` field includes the URL/IP and port that the SDL cloud app can be connected to. 
 
-Endpoints can be added, removed, or updated via policy table updates. This means OEMs will be in control of the web apps that their head units attempt to connect to. 
+The `certificate` field is used for secured RPC service connections which equates to opening a secured websocket connection. 
 
-#### Transport Adapter Connection Flows
+The `enabled` field is by default set to false so the cloud app does not show on the HMI. If the user enables an app via the app store flow highlighted in the section above, then then `enabled` will be set to true in the head units local policy table. When set to true, this cloud app will be included in the HMI RPC "UpdateAppList".
```
theresalech: Please remove extra "then" in second sentence (...then then `enabled` will...)

proposals/0158-cloud-app-transport-adapter.md
```diff
@@ -71,35 +68,133 @@ This is what the app_policies section of a policy table that includes a web app
...
+
+If the websocket connection attempt is successful, Core will send the CloudConnectionStatus enum "CONNECTED" via UpdateAppList.
+
+#### Transport Adapter Connection Flow
+
+App Activation: User selects the cloud app from the app list on the HMI and the websocket connection to the cloud server is opened and auth information is sent.
+
+![alt text](../assets/proposals/0158-cloud-app-transport-adapter/cloud_app_activation.png "App Activation")
+
+The behaviour of displaying a cloud app before registration is to prevent unnecessary websocket connections when an app is not in use. Also if an application is put into hmi status NONE, then the websocket connection will be closed until the user activates the application again.
```
theresalech: Please update to "behavior"

PR #474, opened 2018-04-24
Use Android Annotations Library

bilal-alsharifi:
This proposal is to include the annotations dependency and start using it in the SDL Android library. Java annotations allow developers to provide hints to the compiler and code inspection tools. This ensures that code issues do not propagate which enhances the overall quality of the code.


bilal-alsharifi:

Issue #479, opened 2018-04-25
[Accepted with Revisions] SDL 0165 - Remote Control – Lights – More Names and Status Values


theresalech:
Hello SDL community,

The review of "Remote Control – Lights – More Names and Status Values" begins now and runs through May 1, 2018. The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0165-rc-lights-more-names-and-status-values.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/479

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Theresa Lech

Program Manager - Livio
theresa@livio.io

joeljfischer:
Just a few small notes:

1. I'd like to see documentation on each of the enum names.
2. What error and `info` should be returned if the developer uses an enum they shouldn't? I'd say probably `INVALID_DATA` and `info="The LightStatus enum passed is READ ONLY and cannot be written."

yang1070:
@joeljfischer For your 2nd question, I agree with your suggestion in the case if the application uses a light status enum value that meant to be read-only.  info="The LightStatus enum passed is READ ONLY and cannot be written.", result code `INVALID_DATA` is good for indicates the error, but probably code `READ_ONLY` is more specific.

There is another case,  if the application uses a light name that is not in the supported light name list returned by the capabilities, sdl can use result code `UNSUPPORTED_RESOURCE`, and info "The requested LightName is not supported by the vehicle"

AndrewRMitchell:
I'd also like to see documentation on each of the light names. And I had to go read 099 to get full context to make sure I was commenting on things that already existed (I was about to).

yang1070:
```xml
+   <element name="REAR_CARGO_LIGHTS" value="512">
+     <description>
[...]

PR #487, opened 2018-05-03
Remote Control – Lights – More Names and Status Values

yang1070:
The light status signal in the vehicle can provide more values than just ON / OFF for certain lights. Some vehicles can provide the status of more lights than currently defined light names. This proposal adds some new light status values and some new light names to allow an application to get the status of additional lights.


yang1070:
@theresalech This is the PR to update proposal 165 to reflect the agreed upon revisions here https://github.com/smartdevicelink/sdl_evolution/issues/479. Thank you.

PR #503, opened 2018-05-30
Deprecate RPCRequestFactory

bilal-alsharifi:
This proposal is to deprecate the `RPCRequestFactory` class and start using RPC constructors directly to create new RPC instances.


bilal-alsharifi:

PR #548, opened 2018-07-12
Revise SDL-0150 - Video Streaming State

joeljfischer:
# Revise SDL-0150 - Video Streaming State

## Introduction

Alter `videoStreamingState` to be optional instead of mandatory.

## Motivation

The existing proposal is a major RPC version change and causes high impact backward compatibility issues that are very difficult to mitigate.

## Proposed solution

Change the `videoStreamingState` parameter to optional.

## Potential downsides

## Impact on existing code

Makes a major version change into a minor version change.

## Alternatives considered


joeljfischer:

PR #560, opened 2018-07-23
Revise SDL-0041 - AppIcon Resumption

joeljfischer:
# Revise SDL-0041 - AppIcon Resumption

## Introduction

Alter `iconResumed` to be optional instead of mandatory.

## Motivation

The existing proposal is a major RPC version change without necessity. It causes some oddities within the mobile library regarding `nonnull` and `nullable`.

## Proposed solution

Change the `iconResumed ` parameter to optional. This update also makes some formatting fixes.

## Potential downsides

## Impact on existing code

Makes a major version change into a minor version change.

## Alternatives considered


joeljfischer:
@theresalech please review, thanks!

PR #563, opened 2018-07-25
Revised 0189 Restructuring OnResetTimeout

atiwari9:
Updated the proposal with following changes:
* Replaced _extensionPeriod_ with _resetPeriod_
* Added a note that:
>HMI Integration Guidelines will need to be updated to call out that there is not currently version negotiation between HMI and Core, so older HMI implementations will not work with this new version of Core.


atiwari9:
@theresalech  - I have updated the proposal https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0189-Restructuring-OnResetTimeout.md 

it is ready to be merged.

PR #567, opened 2018-07-30
High level interface: Overlay controllers

kshala-ford:
This proposal is based on [SDL 0156 High level interface: Foundation](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0156-high-level-interface-foundation.md) and [SDL 0176 High level interface: Views and Controllers](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0176-high-level-interface-views-and-controllers.md) adding modal view controllers to the high level interface. These modal view controllers will make the use of RPcs such as `Alert`, `ScrollableMessage` etc. easier and more intuitive to iOS app developers.


kshala-ford:
@theresalech The proposal is ready for review. @joeljfischer your help in reviewing from the techncal side is greatly appreciated. I remember you had some feedback in the old proposal but I don't recall the details...

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
+# High level interface: Modal View Controllers
+
+* Proposal: [SDL-NNNN](NNNN-high-level-interface-modal-view-controllers.md)
+* Author: [Kujtim Shala](https://github.com/kshala-ford)
+* Status: **Awaiting Review**
+* Impacted Platforms: [ iOS ]
+
+## Introduction
+
+This proposal is based on [SDL 0156 High level interface: Foundation](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0156-high-level-interface-foundation.md) and [SDL 0176 High level interface: Views and Controllers](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0176-high-level-interface-views-and-controllers.md)) adding modal view controllers to the high level interface. These modal view controllers will make the use of RPcs such as `Alert`, `ScrollableMessage` etc. easier and more intuitive to iOS app developers.
```
theresalech: Please fix typo: "RPCs"

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+
+This proposal is about adding modal view controllers to the SDL library which extends `SDLViewController` (see [SDL 0176](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0176-high-level-interface-views-and-controllers.md)) used by `SDLApplication` (see [SDL 0156](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0156-high-level-interface-foundation.md)). With modal view controllers app developers can use RPCs showing overlays (such as Alert or Slider) and structure the SDL related code.
+
+This proposal does not contain a modal representation of interacting choice sets. This will be proposed separately.
+
+### Modal view controllers
+
+Modal view controllers are responsible for overlay related RPCs such as `Alert`, `ScrollableMessage`, `PerformAudioPassThru`, `Slider` or `PerformInteraction` (proposed separately). Implementing overlays as modal view controllers would follow the modal-mode presentation of view controllers such as [`UIAlertController`](https://developer.apple.com/documentation/uikit/uialertcontroller).
+
+Apps can instanciate modal view controller classes and present them using the view controller manager (VC manager).
```
theresalech: Please fix typo: "instantiate"

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * Presents a modal view controller on the screen.
+ * Calls the completion handler if the modal view controller disappears from the screen or if an error occurred.
+ */
+- (void)presentModalViewController:(SDLModalViewController *)modalViewController completion:(void (^)(SDLResult result, NSError *error))completion;
+
+@end
+```
+
+When a view controller (presenting VC) want's to present a modal view controller (presented VC) it must call the `presentModalViewController:completion:` method of the view controller manager. The presenting VC must be listed in the view controller stack but does not need to be the top view controller. The presenting VC, presented VC and the completion handler are managed in a VC pair.
```
theresalech: Please fix typo: "wants"

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+
+@end
+```
+
+#### SDLSliderController
+
+This controller matches very closely to the view [`UISlider`](https://developer.apple.com/documentation/uikit/uislider) but for consistency to SDL it makes more sense to treat it as a modal controller. In order to work with the RPC `Slider` it is important to provide the selected value back to the caller. As of UIKit this is not done by using a custom completion handler but simply provide the result in a property of the modal view controller. As per mobile API the property `Slider.sliderFooter` is an array used for two purposes. Either it's a footer (single item) or a list of names representing slider values (number of items must match `.numTicks`). The mobile API allows `.numTicks` to be between 2 and 26. The value range for the parameter `.position` can between 1 and 26 which means the range of values is not more than 26. With the high level abstraction the modal controller can manage the specified range allowing different min and max values. Examples are ranges with negative which are mapped internally to the available range. Future versions of the mobile API could increase the range which would be managed by the slider controller.
```
theresalech: Please update "The value range for the parameter `.position` can between 1 and 26..." to "The value range for the parameter `.position` can be between 1 and 26..." and "Examples are ranges with negative which are mapped..." to "Examples are ranges with negative values..."

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+
+```objc
+@interface SDLSliderController : SDLModalViewController
+
+@property (nonatomic, copy) NSString *title; // mapped with Slider.sliderHeader
+@property (nonatomic, copy) NSArray<NSString *> *valueLables; // overrides control of SDLModalViewController.message
```
theresalech: Should `*valueLables` be `*valueLabels` ?

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+
+// maps to `Slider.position`
+@property (nonatomic) NSInteger value;
+
+@end
+```
+
+## Potential downsides
+
+The initial workload in order to implement this high level interface is expected to be quite high. Once implemented it is expected that developers will be able to implement SDL into their apps in less time than they would need today. At the end the maintenance of the high level interface may be lower compared to the counterproposal for different reasons.
```
theresalech: Can you include a reference to the counterproposal here?

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+#### SDLScrollableMessageController
+
+Similar to `Alert` this view controller takes care of the RPC `ScrollableMessage`.
+
+```objc
+@interface SDLScrollableMessageController : SDLModalViewController
+
+@property (nonatomic, copy) NSArray<SDLSoftButton *> *buttons;
```
joeljfischer: Needs a body property

joeljfischer: These should probably be readonly and created in the setter. If they change them during the presentation, things could get weird or unexpected for the developer

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+
+@property (nonatomic, copy) NSString *title; // mapped with Slider.sliderHeader
+@property (nonatomic, copy) NSArray<NSString *> *valueLables; // overrides control of SDLModalViewController.message 
+
+// minimumValue and maximumValue map to `Slider.numTicks`
+@property (nonatomic) NSInteger minimumValue;
+@property (nonatomic) NSInteger maximumValue;
+
+// maps to `Slider.position`
+@property (nonatomic) NSInteger value;
```
joeljfischer: `initialPosition`?

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+
+```objc
+@interface SDLSliderController : SDLModalViewController
+
+@property (nonatomic, copy) NSString *title; // mapped with Slider.sliderHeader
+@property (nonatomic, copy) NSArray<NSString *> *valueLables; // overrides control of SDLModalViewController.message 
+
+// minimumValue and maximumValue map to `Slider.numTicks`
+@property (nonatomic) NSInteger minimumValue;
+@property (nonatomic) NSInteger maximumValue;
```
joeljfischer: I don't think this is how that works, e.g. if you get `10` and `11`, you'll actually get 2. This would have to map to the value labels. Can't we just get the number of ticks from the `valueLabels` above?

proposals/NNNN-high-level-interface-modal-view-controllers.md
joeljfischer: `labels`?

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+
+```objc
+@interface SDLSliderController : SDLModalViewController
+
+@property (nonatomic, copy) NSString *title; // mapped with Slider.sliderHeader
```
joeljfischer: Missing a footer property

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+
+#### SDLViewControllerManager (additions)
+
+```objc
+@interface SDLViewControllerManager
+
+/**
+ * This view controller is currently presenting a modal view on the screen.
+ */
+@property (nonatomic) SDLViewController *presentingModalViewController;
```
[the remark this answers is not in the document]
kshala-ford: Actually the idea is to not limit this to the top VC. Same applies to UIKit (not limiting to top VC either). See line 57:

> The presenting VC must be listed in the view controller stack but does not need to be the top view controller.

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+
+/**
+ * This modal view controller is currently presented on the screen (by the presenting view controller).
+ */
+@property (nonatomic) SDLModalViewController *presentedModalViewController;
```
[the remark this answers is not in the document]
kshala-ford: I wanted to be clear with property naming. The presented VC is a modal VC not a regular VC.

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+
+/**
+ * Presents a modal view controller on the screen.
+ * Calls the completion handler if the modal view controller disappears from the screen or if an error occurred.
+ */
+- (void)presentModalViewController:(SDLModalViewController *)modalViewController completion:(void (^)(SDLResult result, NSError *error))completion;
```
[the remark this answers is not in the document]
kshala-ford: `CPInterfaceController.presentTemplate` is for non-modal UI. The API is more aligned with UIKit than with CarPlay.

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+The alert controller is used to abstract the RPC `Alert`. The properties match the parameters of the RPC.
+
+```objc
+@interface SDLAlertController : SDLModalViewController
+
+@property (nonatomic, copy) NSArray<SDLTTSChunk *> *initialPrompt;
+@property (nonatomic, copy) NSArray<SDLSoftButton *> *buttons;
+
+@property (nonatomic) BOOL shouldPlayTone;
+@property (nonatomic) BOOL progressIndicator;
```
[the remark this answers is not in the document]
kshala-ford: good point. `shouldShowProgressIndicator` (from API naming "show" appears more often)

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+The audio input controller is used to abstract the RPC `PerformAudioPassThru`, `OnAudioPassThru` and `EndAudioPassThru`. The app developer should be able to provide a block which is called for audio data. Furthermore a method called `dismiss` should help to end the audio pass thru.
+
+```objc
+@interface SDLAudioInputController : SDLModalViewController
+
+@property (nonatomic, copy) NSArray<SDLTTSChunk *> *initialPrompt;
+@property (nonatomic, copy) SDLSamplingRate samplingRate;
+@property (nonatomic, copy) SDLBitsPerSample bitsPerSample;
+@property (nonatomic, copy) SDLAudioType audioType;
+@property (nonatomic) BOOL muteMediaSource;
```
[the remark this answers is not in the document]
kshala-ford: The changes would simply not be applied during presentation but you're right. I'll make them (and from other classes) read only including a dedicated init.

proposals/NNNN-high-level-interface-modal-view-controllers.md
kshala-ford: Actually it doesn't need a body. `SDLModalViewController.message` will be used.

proposals/NNNN-high-level-interface-modal-view-controllers.md
kshala-ford: different to other properties `.value` should not be read-only. It should cover the request's param `.position` as well as the response's param `.sliderPosition`. Needs documentation that programmatically changing the value during presentation will be ignored/not be applied.

proposals/NNNN-high-level-interface-modal-view-controllers.md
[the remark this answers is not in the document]
joeljfischer: That's not really true though. In later versions of iOS that functionality was deprecated and modal VCs have to be presented from the top view controller. That's why the iOS SDL library's lock screen, for example, (before it was a separate `UIWindow`) had code to traverse the navigation hierarchy and find the top-most VC in order to present the modal VC on top of that.

proposals/NNNN-high-level-interface-modal-view-controllers.md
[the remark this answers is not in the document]
joeljfischer: The type of the property should be enough to make that clear. I'm okay with the naming if you're strongly against a change, it just seems a bit unnecessarily long.

Also, this should be readonly.

proposals/NNNN-high-level-interface-modal-view-controllers.md
kshala-ford: I should have mentioned that the footer is covered by the modalVC.message property and ignored if valueLabels is used.

proposals/NNNN-high-level-interface-modal-view-controllers.md
kshala-ford: `labels`! ^^

proposals/NNNN-high-level-interface-modal-view-controllers.md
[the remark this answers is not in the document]
joeljfischer: I believe it's actually only for modal UI. They have `pushTemplate`, and `popTemplate` for the non-modal case. What about `presentModal`?

Additionally, there has been discussion on the SDLC steering committee calls about explicitly not conforming these types of what we call modal behaviors as modals. During core certification guideline review, mention of "pop-up" and "modal" was explicitly criticized and removed due to some steering committee members not wanting to be restricted into making those objects modal. I'm wondering if we should rename this, but I'm not sure what it would be renamed to. Perhaps these could simply be `SDLView`s and the primary layouts `SDLTemplate`s?

proposals/NNNN-high-level-interface-modal-view-controllers.md
joeljfischer: I think that was a criticism of mine on the first proposal. If additional modals are created that _don't_ have a message, then having a `message` property on `SDLModalViewController` becomes a problem. I am personally strongly in favor of removing `message` from there and putting it on each subclass.

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+
+#### SDLAlertController
+
+The alert controller is used to abstract the RPC `Alert`. The properties match the parameters of the RPC.
+
+```objc
+@interface SDLAlertController : SDLModalViewController
+
+@property (nonatomic, copy) NSArray<SDLTTSChunk *> *initialPrompt;
+@property (nonatomic, copy) NSArray<SDLSoftButton *> *buttons;
```
joeljfischer: This should probably be `SDLSoftButtonObject` so that developers don't have to deal with `buttonId` and can auto-upload an image?

proposals/NNNN-high-level-interface-modal-view-controllers.md
joeljfischer: How does the VC get notified that it's done? It's through the completion handler on the VC manager? Then the property gets re-inspected? I can see this being confusing for developers, even though it's similar to how `UISlider` works, because it's still significantly different due to differences between UIKit and SDL (the UIKit view stays on screen vs. the SDL is modal only and will disappear, the value changes as the user changes it vs. the value is only updated when the view disappears). Unless there's a special initializer that can provide a callback when the value changes (e.g. on the response)? I'm not sure of the best way to approach it.

proposals/NNNN-high-level-interface-modal-view-controllers.md
joeljfischer: Even still, I don't think `min` and `max` map correctly. As I said above, if you set `minimum` to 11 and `maximum` to 12, the value you get back is either 1 or 2 (unless you're saying we're interpolating back into that range)? Additionally, there are no limits here despite that we're limited to 26 maximum values (side-note, that should probably be expanded).

Perhaps it should be two initializers, one for `valueLabels` and one for `numValues`? But only expose `valueLabels` with empty strings if no labels were set?

proposals/NNNN-high-level-interface-modal-view-controllers.md
joeljfischer: I think that's pretty confusing for developers. See my comment above about splitting `message` into each subclass. I think that this is a good case where we're going to cause developer confusion by having `message` in the subclass. I would never have guessed that if you hadn't told me.

proposals/NNNN-high-level-interface-modal-view-controllers.md
```diff
@@ -0,0 +1,160 @@
...
+```objc
+@interface SDLAudioInputController : SDLModalViewController
+
+@property (nonatomic, copy) NSArray<SDLTTSChunk *> *initialPrompt;
+@property (nonatomic, copy) SDLSamplingRate samplingRate;
+@property (nonatomic, copy) SDLBitsPerSample bitsPerSample;
+@property (nonatomic, copy) SDLAudioType audioType;
+@property (nonatomic) BOOL muteMediaSource;
+
+@property (nonatomic) (^onAudioData)(NSData *audioData);
```
joeljfischer: `onAudioData` seems perhaps a bit out of place compared to the rest of this API and iOS' naming.

proposals/NNNN-high-level-interface-overlay-controllers.md
```diff
@@ -0,0 +1,326 @@
...
+- the use of RPCs and handle notifications, requests and responses
+- manually manage concurrent and sequential operations (e.g. image upload)
+
+The time for an app developer to learn SDL (non-productive time) is high and not accepted by some developers. The SDL iOS library already improves the current situation by abstracting painful implementations. However the management layer is still unfamiliar and causes a high learning curve.
+
+## Proposed solution
+
+This proposal is about adding overlay controllers to the SDL library which extends `SDLViewController` (see [SDL 0176](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0176-high-level-interface-views-and-controllers.md)) used by `SDLApplication` (see [SDL 0156](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0156-high-level-interface-foundation.md)). With overlay controllers app developers can use RPCs such as Alert, ScrollableMessage or Slider and structure the SDL related code.
+
+This proposal does not contain a overlay controller for choice sets. This will be proposed separately.
```
theresalech: Please update to "...contain an overlay controller..."

proposals/NNNN-high-level-interface-overlay-controllers.md
```diff
@@ -0,0 +1,326 @@
...
+- read best practices and implement proxy (manager) lifecycle
+- the use of RPCs and handle notifications, requests and responses
+- manually manage concurrent and sequential operations (e.g. image upload)
+
+The time for an app developer to learn SDL (non-productive time) is high and not accepted by some developers. The SDL iOS library already improves the current situation by abstracting painful implementations. However the management layer is still unfamiliar and causes a high learning curve.
+
+## Proposed solution
+
+This proposal is about adding overlay controllers to the SDL library which extends `SDLViewController` (see [SDL 0176](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0176-high-level-interface-views-and-controllers.md)) used by `SDLApplication` (see [SDL 0156](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0156-high-level-interface-foundation.md)). With overlay controllers app developers can use RPCs such as Alert, ScrollableMessage or Slider and structure the SDL related code.
```
theresalech: Please add code blocks around `Alert`, `ScrollableMessage`, and `Slider`

proposals/NNNN-high-level-interface-overlay-controllers.md
```diff
@@ -0,0 +1,326 @@
...
+@property (nonatomic, strong) SDLOverlayController *overlayController;
+
+@property (nonatomic, copy) SDLOverlayControllerCompletionHandler completionHandler;
+
+@end
+```
+
+When a view controller wants to present an overlay controller it must call the `presentOverlayController:completion:` method of the VC manager. 
+The presenting VC must be listed in the view controller stack but does not need to be the top view controller. The VC manager should initiate the presentation 
+(call `present` of the overlay controller class) and store the overlay controller and completion handler in an internal list. The overlay controller should be able to listen to responses and notification relevant to the presentation. If the overlay controller has detected the end of the presentation it should notify the VC manager. This can also be done through an encapsulated overlay controller completion handler.
```
theresalech: I think this should say "...listen to responses and notifications relevant..."

kshala-ford:
@theresalech thank you for the review. everything should be fixed

PR #572, opened 2018-08-01
Revise SDL-0099 - New remote control modules

joeljfischer:
## Introduction

Remove `GPSLocation` struct and reuse `GPSData`, which as of [SDL-0175](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0175-Updating-DOP-value-range-for-GPS-notification.md) supports optional parameters. Additionally, remove `sRGBColor` struct and reuse `RGBColor` from [SDL-0147](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0147-template-color-scheme.md).

## Motivation

The existing proposal adds duplicate structs.

## Proposed solution

Remove the duplicate structs and reuse `GPSData` and `RGBColor`.

## Potential downsides

## Impact on existing code

There will be less duplicated code.

## Alternatives considered

### DIFF
--- a/proposals/0099-new-remote-control-modules-and-parameters.md
+++ b/proposals/0099-new-remote-control-modules-and-parameters.md
@@ -3,7 +3,7 @@
 * Proposal: [SDL-0099](0099-new-remote-control-modules-and-parameters.md)
 * Author: [Zhimin Yang](https://github.com/yang1070)
 * Status: **Accepted with Revisions**
-* Impacted Platforms: [Core / iOS / Android / RPC ]
+* Impacted Platforms: [Core / iOS / Android / RPC]
 
 ## Introduction
 
@@ -201,15 +201,6 @@ Add new a parameter to RADIO.
     </param>
   </struct>
 
-  <struct name="GPSLocation">
-    <param name="longitudeDegrees" type="Float" minvalue="-180" maxvalue="180" mandatory="true">
-    </param>
-    <param name="latitudeDegrees" type="Float" minvalue="-90" maxvalue="90" mandatory="true">
-    </param>
-    <param name="altitudeMeters" type="Float" minvalue="-99999" maxvalue="99999" mandatory="false">
-    </param>
-  </struct>
-
   <struct name="SisData">
     <param name="stationShortName" type="String" minlength="4" maxlength="7"  mandatory="false">
       <description>Identifies the 4-alpha-character station call sign plus an optional (-FM) extension</description>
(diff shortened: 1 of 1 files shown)


joeljfischer:
@theresalech all done

yang1070:
This will be a breaking change as reusing the data structure will not only change the core internal implementation but also the definitions in the mobile_api.xml and hmi_api.xml. Both core and mobile proxy libraries need update to use the new data structure.

stefanek22:
Can we align on a timing plan to update and deliver any associated code changes for both SDL Core and the Mobile SDK?  This has impact on the existing deliveries for the upcoming SDL Core 5.0.0 release.

joeljfischer:
Hi @yang1070 this is not a breaking change. This removes RPCs _before_ they are released, therefore this is not any type of an API change.

joeljfischer:
@stefanek22 For the mobile SDK, or at least the iOS SDK, the changes would be minimal (less than an hour). Core is the big question here, and hopefully @JackLivio can shed some light on how long this would take to change there?

joeljfischer:
I just noticed I named an RPC parameter `RGBColorSpaceAvailable`, when it should be `rgbColorSpaceAvailable`.

JackLivio:
@joeljfischer We haven't reviewed the original Core implementation yet for these new RC modules. I imagine the required changes would be small. No logic changes needed, only name changes.

yang1070:
@joeljfischer I agree to reuse the existing data struct whenever it is possible, There shall be no issue regarding the GRBColor, but there is an issue regarding the GPSLocation data structure. According to the mobile_api.xml and hmi_api.xml in the develop branch and the update of GPS data PR https://github.com/smartdevicelink/sdl_core/pull/2417, even after the change, there are still many mandatory fields (not available in the RC use case, the location of radio station comes from the radio broadcast signals, not the GPS device) required in the data structure. RC cannot use it unless the proposal changes all other fields to be not mandatory. And that change will be breaking change.

joeljfischer:
Should the HMI API for SDL-0175 also have been updated and it was a mistake that the Mobile API was the only part updated? We should not use the `GPSLocation` structure if we can manage it, the API duplication is simply too egregious.

We are already doing a breaking change to the HMI API in Core 5.0, are we not? So this would be our best opportunity to do such an update?

yang1070:
@joeljfischer  Yes, I agree that SDL-0175 should make the other fields optimal in `GPSData` in the HMI API.

yang1070:
It is strange in master branch, every thing of `GPSData` is not mandatory, but in develop branch some are mandatory.

yang1070:
@joeljfischer I have create another PR https://github.com/smartdevicelink/sdl_evolution/pull/577 for this

joeljfischer:
@yang1070 👍🏻

PR #575, opened 2018-08-02
Revised Retry Failed File Uploads

NicoleYarroch:
Revised the retry failed file uploads proposal to include Android as an impacted platform


NicoleYarroch:
@theresalech I made the revisions to SDL 0191.

PR #578, opened 2018-08-06
Revise SDL-0153 Support for Short and Full UUID App ID

JackLivio:
## Introduction
This is a revision to make fullAppID mandatory="false" and to add a clarifying note regarding appID case insensitivity.

#### Make Full App ID Non Mandatory
There is language in the original proposal conflicting with the fullAppID RPC spec definition.

From proposal:
> If only the (short) appID is present in the RegisterAppInterface RPC request (legacy apps), sdl core shall use the (short) appID as the internal policy app id. i.e. use appID as the policy_app_id and use it to construct the icon path.

In order for this case to be possible, fullAppID must be non mandatory. In addition, even with mobile version negotiation, making fullAppID mandatory would cause a breaking change in RegisterAppInterfaceRequest for legacy applications.

#### Include Note Regarding AppID Case Insensitivity
During the Core implementation of this original proposal, it was discovered that there was an existing behavior in Core that converts incoming appIDs from a RegisterAppInterfaceRequest to lowercase. I am adding this note to the proposal for documentation and visibility purposes. 

## Proposed solution

Change fullAppID rpc spec signature to: 

| Value |  Type | Mandatory | Description | 
| ---------- | ---------- |:-----------: |:-----------:|
|`fullAppID`|String|False|ID used to validate app with policy table entries|

## Potential downsides

## Impact on existing code
Change mandatory value `fullAppID` in mobile_api.xml in RPC Spec to false.

## Alternatives considered

### DIFF
--- a/proposals/0153-support-short-long-appid.md
+++ b/proposals/0153-support-short-long-appid.md
@@ -21,7 +21,7 @@ Add a new variable to RegisterAppInterface request:
 
 | Value |  Type | Mandatory | Description | 
 | ---------- | ---------- |:-----------: |:-----------:|
-|`fullAppID`|String|True|ID used to validate app with policy table entries|
+|`fullAppID`|String|False|ID used to validate app with policy table entries|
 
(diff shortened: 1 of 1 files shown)


JackLivio:
@theresalech Ready for review

theresalech:
@JackLivio can you please update the description to include the proper proposal formatting (introduction, motivation, etc.)?  Here's an example for reference: https://github.com/smartdevicelink/sdl_evolution/pull/572.  Let me know of any questions, thanks!

PR #592, opened 2018-08-24
Update SetMediaClockTimer Initializers

joeljfischer:
This proposal is to update the SDL mobile library `SetMediaClockTimer` class with initializers to assist developers with creating RPCs for each of the use cases of that RPC.


joeljfischer:
@theresalech ready for review, thank you!

PR #597, opened 2018-09-12
Adding GPS Shift support

atiwari9:
To add _shifted_ param for GPS data struct so that HMI can communicate to app when GPS data is shifted using an algorithm


atiwari9:
@theresalech  - This proposal is ready to be merged

PR #621, opened 2018-11-21
Support running the same app from multiple devices at the same time

yang1070:
This proposal proposes that SDL shall support running the same app from multiple devices at the same time. Currently, SDL only allows one instance of the same app to register with SDL.  However, there are valid use cases for running the same app from two or more devices at the same time. 

For example, two `Pandora` apps are running on an iPhone and an android phone respectively. Both devices connected to the vehicle’s infotainment system. They are streaming music to front and rear audio respectively.

As another example, two climate-remote-control apps are running on two devices that both connected to the vehicle. One used by the front passenger, the other used by a passenger sitting in the rear seat. They control the front and rear climate features separately.

### DIFF
--- a/proposals/0204-same-app-from-multiple-devices.md
+++ b/proposals/0204-same-app-from-multiple-devices.md
@@ -15,8 +15,8 @@ As another example, two climate-remote-control apps are running on two devices t
 
 ## Motivation
 
-SDL supports the following transport types, Bluetooth, USB-AOA, and TCP (WiFi) out of the box. OEMs can also extend that list to include iAP over BT or iAP over USB. For details, please see link 
-https://smartdevicelink.com/en/docs/hmi/master/basiccommunication/updatedevicelist/ . It is possible that multiple devices connect to SDL at the same time using either the same (TCP) or different transports. 
+SDL supports the following transport types, Bluetooth, USB-AOA, and TCP (WiFi) out of the box. OEMs can also extend that list to include iAP over BT or iAP over USB. For details, please see [link here](
+https://smartdevicelink.com/en/docs/hmi/master/basiccommunication/updatedevicelist/). It is possible that multiple devices connect to SDL at the same time using either the same (TCP) or different transports. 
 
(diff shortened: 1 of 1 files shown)


yang1070:
@jordynmackool proposal is ready for review

proposals/0204-same-app-from-multiple-devices.md
```diff
@@ -43,14 +43,40 @@ There is no API change. Just updates some descriptions.
...
 +        Additional applications with the same name from the same device will be rejected.
        Only characters from char set [@TODO: Create char set (character/hex value) for each ACM and refer to] are supported.
     </description>
 </param>
 ```
 
+```xml
+<param name="ttsName" type="TTSChunk" minsize="1" maxsize="100" array="true" mandatory="false" since="2.0">
+    <description>
+        TTS string for VR recognition of the mobile application name, e.g. "Ford Drive Green".
```
JackLivio: Please remove Ford specific references from API.

proposals/0204-same-app-from-multiple-devices.md
```diff
@@ -43,14 +43,40 @@ There is no API change. Just updates some descriptions.
...
+        Only characters from char set [@TODO: Create char set (character/hex value) for each ACM and refer to] are supported.
+    </description>
+</param>
+```
+
+```xml
+<param name="vrSynonyms" type="String" maxlength="40" minsize="1" maxsize="100" array="true" mandatory="false" since="1.0">
+    <description>
+        Defines an additional voice recognition command.
+-        May not interfere with any app name of previously registered applications and any predefined blacklist of words (global commands)
```
JackLivio: Remove this line?

JackLivio: I don't think the diff merge worked properly

JackLivio:
I think there was a problem merging in these changes? I am seeing a lot of "++", "+-", "--", "-+" on the line changes.

proposals/0204-same-app-from-multiple-devices.md
yang1070: The string has been already in the xml API file. I didn't introduce it in the proposal.

proposals/0204-same-app-from-multiple-devices.md
yang1070: keep these lines. The intention is when we update the API xml later, the line shall be removed/updated.

PR #624, opened 2018-11-26
Block Old SDL Head Units from Mobile

joeljfischer:
Add a way for proxy apps to block older head units from registering and therefore to not appear on those head units.


joeljfischer:
@jordynmackool This is ready for review

joeljfischer:
@jordynmackool Thanks for the review, I have updated the proposal

PR #626, opened 2018-11-27
ATF build using CMake

LuxoftAKutsan:
This proposal offers using CMake build system for compilation ATF project instead of qmake.


lauratonwe:
We are closing this and we resubmit after revisions and Ford review

PR #656, opened 2019-01-02
iOS ReplayKit Streaming

SatbirTanda:
This proposal aims to optimize how the car window API captures the screen in the iOS Framework using ReplayKit.

### DIFF
--- a/proposals/0215-ios-replaykit-streaming.md
+++ b/proposals/0215-ios-replaykit-streaming.md
@@ -1,0 +1,95 @@
+# iOS ReplayKit Streaming
+
+* Proposal: [SDL-0215](0215-ios-replaykit-streaming.md)
+* Author: [Satbir Tanda](https://github.com/satbirtanda)
+* Status: **Awaiting review**
+* Impacted Platforms: [iOS]
+
+## Introduction
+
+This proposal aims to optimize how the car window API captures the screen in the iOS Framework using ReplayKit.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


SatbirTanda:
@jordynmackool The PR is ready for review.

proposals/0215-ios-replaykit-streaming.md
```diff
@@ -0,0 +1,95 @@
...
+## Potential downsides
+
+- ReplayKit displays an alert asking the user to allow the app to record the screen.  However, after the user taps this once, it will only ask again on a new app launch after a certain time window
+- The app would need to support horizontal orientation otherwise the screen would be captured in portrait
+- Must be on iOS 11 or greater
+
+## Impact on existing code
+
+* No changes to the Android library
+* The iOS library would need to alter it's code to add a flag to pick between ReplayKit streaming or the current method of streaming (possibly via the ```SDLStreamingMediaConfiguration``` ).  Also it would need to alter SDLCarWindow.m's ```syncFrame``` method that causes the jump in CPU usage (which can be seen via the CPU profiling instrument while running an SDL enabled navigation app by filtering out system methods).  Also the app would need to be forced in horizontal orientation.
```
jordynmackool: Update to:
 The iOS library would need to alter it's code to add a flag to pick between ReplayKit streaming or the current method of streaming (possibly via the ```SDLStreamingMediaConfiguration``` ).  **Additionally, it** would need to alter SDLCarWindow.m's ```syncFrame``` method that causes the jump in CPU usage (which can be seen via the CPU profiling instrument while running an SDL enabled navigation app by filtering out system methods).  **The app** would need to be forced **in a horizontal** orientation.

Issue #606, opened 2018-10-17
[Returned for Revisions] SDL 0201 - High level interface: Overlay Controllers


jordynmackool:
Hello SDL community,

The review of "SDL 0201 - High level interface: Overlay Controllers" begins now and runs through February 19, 2019. The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0201-high-level-interface-overlay-controllers.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/606

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Jordyn Mackool

Program Manager - Livio
jordyn@livio.io

joeljfischer:
Hi, I'm happy to see this has moved to review. There's much that I agree with here, but I do have a number of notes:

1. The name of "overlay controller" may be misleading. When doing the review of the Core Certification guidelines, it was argued that some of these operations may not be done in an overlay style like Ford's Gen 3 (specifically, I think AudioPassThru was called out, as it's difficult to imagine others occurring without an overlay, though they could replace the template wholesale as well). I believe I originally criticized the name "Modal" for this very reason, however, I think it's better than "Overlay," and would prefer it if nothing better can be provided. At least "Modal" is describing various "modes of operation" rather than a UI overlay, which may or may not happen based on the OEM implementation.

2. I'm not sure that storing pending overlay controllers is better than just failing when one is already presented. I can see both being useful at different times, and the dev could implement either manually if they wanted it. Personally, I think I'd prefer failing. 🤷‍♂️  I've re-read the paragraph describing this a few times, but I think I'm still missing the purpose of allowing it, could you elaborate?

[...]

PR #659, opened 2019-01-14
High Level Interface: Overlay Controller (revision)

kshala-ford:
This PR is related to "SDL 0201 - High level interface: Overlay Controllers" #606 and contains revisions requested in https://github.com/smartdevicelink/sdl_evolution/issues/606#issuecomment-448346334

### DIFF
--- a/proposals/0201-high-level-interface-overlay-controllers.md
+++ b/proposals/0201-high-level-interface-overlay-controllers.md
@@ -154,8 +154,9 @@ To do that the app developer must create a new view and place it in the `view` p
 @interface SDLView
 
 /**
- * Specifies wether the view is opaque. An opaque view will be presented on the app's screen overriding content from underlying views.
- * If this property is set to NO the view is transparent and the view controller manager will composite content from underlying views.
+ * Specifies wether the view is opaque. An opaque view will be presented on the app's screen
+ * overriding all content from underlying views. If this property is set to NO the view is
+ * transparent and the view controller manager will composite content from underlying views.
  * The default value is YES.
  */
 @property (nonatomic, assign) BOOL opaque;
@@ -163,7 +164,7 @@ To do that the app developer must create a new view and place it in the `view` p
 @end
 ```
 
-This `opaque` flag should allow views from the underlying view controller to stay visible while the overlay controller is presenting. The opaque mode is view type related. With this flag 
+If the `opaque` flag is set to NO the view will be called transparent. It should allow views or view parts from the underlying view controller to stay visible while the overlay controller is presenting. The opaque mode is view type related and allows . With this flag 
 - the app developer has full control to decide on what to show and how to show
 - the complexity is scalable per app developer choice (no overlay view -> no screen changes)
 
--- a/proposals/0201-high-level-interface-overlay-controllers.md
+++ b/proposals/0201-high-level-interface-overlay-controllers.md
@@ -156,7 +156,7 @@ Depending on the subclass (and the min/max values of the parameter) the duration
 
 This feature is planned in the proposal for Overlay controller related to choice sets.
 
-To realize above behavior, the app developer must create a new view and place it in the `view` property of the overlay controller. When presenting the overlay controller with a view object the view controller manager would override the full screen of the app but it may be handy to just override a single text field. With that a new property should be added to `SDLView` for an opaque mode.
+To realize the above behavior, the app developer must create a new view and place it in the `view` property of the overlay controller. When presenting the overlay controller with a view object the view controller manager would override the full screen of the app, but it may be handy to just override a single text field. With that a new property should be added to `SDLView` for an opaque mode.
 
 ```objc
 @interface SDLView
@@ -172,7 +172,7 @@ To realize above behavior, the app developer must create a new view and place it
 @end
 ```
 
-If the `opaque` flag is set to NO the view will be called transparent. It should allow views or view parts from the underlying view controller to stay visible while the overlay controller is presenting. The opaque mode is view type related and allows . With this flag:
+If the `opaque` flag is set to NO the view will be called transparent. It should allow views or view parts from the underlying view controller to stay visible while the overlay controller is presenting. The opaque mode is view type related and allows overriding content per view. As an example a transparend overlay, which contains a graphic will only override the graphic view of the underlying view. With this flag:
 - the app developer has full control to decide on what to show and how to show it.
 - the complexity is scalable per the app developers choice (no overlay view -> no screen changes)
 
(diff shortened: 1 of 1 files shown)


kshala-ford:
@jordynmackool The PR is ready for review

kshala-ford:
@jordynmackool ping

proposals/0201-high-level-interface-overlay-controllers.md
```diff
@@ -394,3 +430,26 @@ This proposal will add a total new high level interface layer abstracting many p
 ## Alternatives considered
 
 As discussed in the steering committee meeting from March 20 (see [here](https://github.com/smartdevicelink/sdl_evolution/issues/379#issuecomment-374736496)) this proposal is a counterproposal to [0133 - Enhanced iOS Proxy Interface](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0133-EnhancediOSProxyInterface.md).
+
+Regarding `SDLSliderController` there is a different behavior depending on the existance of a message string. In the main proposal it is proposed to show either value strings *or* the message. The alternative solution would be to alwas show value strings and concatenate the message into each value string if provided. The concatenation would be `@"%ld\n%@"` where `%ld` is the value and `%@` is the message. The upside is that we can merge translated values including a message. The downside is that the message is repeated for each value. Example:
```
jordynmackool: Update to:
Regarding `SDLSliderController` there is a different behavior depending on the **existence** of a message string. In the main proposal it is proposed to show either value strings *or* the message. The alternative solution would be to **always** show value strings and concatenate the message into each value string if provided. The concatenation would be `@"%ld\n%@"` where `%ld` is the value and `%@` is the message. The upside is that we can merge translated values including a message. The downside is that the message is repeated for each value. Example:

kshala-ford:
@jordynmackool I addressed all items. Some I restructured in hope of improved readability.

proposals/0201-high-level-interface-overlay-controllers.md
```diff
@@ -139,15 +142,58 @@ typedef void (^SDLOverlayControllerCompletionHandler)(SDLResult result, NSError
...
-#### Main screen overlay
+### Main screen overlay
+
+`SDLOverlayController` inherits `SDLViewController`. With this, it is possible to manipulate the app's main screen during a presentation. This can be very useful to:
+- show a text message for "Loading..." or "Please wait" if the overlay takes more time to load
+- add buttons to cancel the overlay request if possible.
+
+This feature is planned in the proposal for Overlay controller related to choice sets.
+
+To realize above behavior, the app developer must create a new view and place it in the `view` property of the overlay controller. When presenting the overlay controller with a view object the view controller manager would override the full screen of the app but it may be handy to just override a single text field. With that a new property should be added to `SDLView` for an opaque mode.
```
jordynmackool: Update to:
To realize **the** above behavior, the app developer must create a new view and place it in the `view` property of the overlay controller. When presenting the overlay controller with a view object the view controller manager would override the full screen of the **app, but** it may be handy to just override a single text field. With that a new property should be added to `SDLView` for an opaque mode.

proposals/0201-high-level-interface-overlay-controllers.md
```diff
@@ -139,15 +142,58 @@ typedef void (^SDLOverlayControllerCompletionHandler)(SDLResult result, NSError
...
+ * overriding all content from underlying views. If this property is set to NO the view is
+ * transparent and the view controller manager will composite content from underlying views.
+ * The default value is YES.
+ */
+@property (nonatomic, assign) BOOL opaque;
+
+@end
+```
+
+If the `opaque` flag is set to NO the view will be called transparent. It should allow views or view parts from the underlying view controller to stay visible while the overlay controller is presenting. The opaque mode is view type related and allows . With this flag:
```
jordynmackool: Is there additional information missing here: `type related and allows .` ?

Update to:
If the `opaque` flag is set to NO the view will be called transparent. It should allow views or view parts from the underlying view controller to stay visible while the overlay controller is presenting. The opaque mode **is a** view **type related and allows .** With this flag:

proposals/0201-high-level-interface-overlay-controllers.md
```diff
@@ -385,6 +428,8 @@ The property `value` maps to `Slider.value` *and*  to `SliderResponse.sliderValu
...
...IAlertController` using local variables. See [`UIViewController.presentViewController:animated:completion:`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621380-presentviewcontroller) and [Objective-C Working with Blocks](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithBlocks/WorkingwithBlocks.html).  The reason for not returning the presented overlay in the block is to avoid the pain of dealing with `__kindof` and casting back from the base class to the sub class. With that, a potential downside may happen if an app developer stores the presented view controller in a strong property which may cause a temporary retain cycle to `self` during presentation. This temporary retain cycle is acceptable as it is automatically released, therefore it will not cause any problems within the app. The gain of avoiding casts and `__kindof` for *every* usage is higher than the rare and hypothetical case of a developer storing the overlay instance in a strong property, especially as it is automatically released after the end of the presentation or after a disconnect. Last The proposed way follows existing practice.
```
jordynmackool: **Last, the** proposed way follows **the** existing practice.

proposals/0201-high-level-interface-overlay-controllers.md
```diff
@@ -394,3 +439,26 @@ This proposal will add a total new high level interface layer abstracting many p
 ## Alternatives considered
 
 As discussed in the steering committee meeting from March 20 (see [here](https://github.com/smartdevicelink/sdl_evolution/issues/379#issuecomment-374736496)) this proposal is a counterproposal to [0133 - Enhanced iOS Proxy Interface](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0133-EnhancediOSProxyInterface.md).
+
+Regarding `SDLSliderController` there is a different behavior depending on the existance of a message string. In the main proposal it is proposed to show either value strings *or* the message. The alternative solution would be to always show value strings and concatenate the message into each value string if provided. The concatenation would be `@"%ld\n%@"` where `%ld` is the value and `%@` is the message. The upside is that we can merge translated values including a message. The downside is that the message is repeated for each value. Example:
```
jordynmackool: Please refrain from using we. Apologies I missed this in the first review.

`We` to `SDLC`

proposals/0201-high-level-interface-overlay-controllers.md
kshala-ford: > Is there additional information missing here: `type related and allows .` ?

oh wow... sorry for that

kshala-ford:
@jordynmackool thanks for reviewing again. I just addressed the last points you mentioned

proposals/0201-high-level-interface-overlay-controllers.md
```diff
@@ -172,7 +172,7 @@ To realize above behavior, the app developer must create a new view and place it
 @end
 ```
 
-If the `opaque` flag is set to NO the view will be called transparent. It should allow views or view parts from the underlying view controller to stay visible while the overlay controller is presenting. The opaque mode is view type related and allows . With this flag:
+If the `opaque` flag is set to NO the view will be called transparent. It should allow views or view parts from the underlying view controller to stay visible while the overlay controller is presenting. The opaque mode is view type related and allows overriding content per view. As an example a transparend overlay, which contains a graphic will only override the graphic view of the underlying view. With this flag:
```
jordynmackool: Update to:
If the `opaque` flag is set to NO the view will be called transparent. It should allow views or view parts from the underlying view controller to stay visible while the overlay controller is presenting. The opaque mode is view type related and allows overriding content per view. **As an example,** a **transparent** overlay contains a graphic that will only override the graphic view of the underlying view. With this flag:

kshala-ford:
@jordynmackool fixed the transparend ^^

Issue #675, opened 2019-02-06
Revisions to 0167-app-services 


jordynmackool:
Hello SDL community,

The original proposal "SDL 0167 - App Services" was accepted by the Steering Committee on 2018-05-23, alterations to the proposal have been requested through a pull request.

This review will focus on only the changes present in this pull request "Revisions to 0167-app-services #673" begins now and runs through February 12, 2019.

**Note: All comments should be contained in this issue!**

The original proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0167-app-services.md

**Noted Revisions**

- Fix various typos and unify naming conventions.
- Make AppServiceType a documentation item and change all uses of the AppServiceType to String. This will allow for support of future app services that core does not recognize.
- Add notes regarding system capability updates.
- Change serviceIcon type to Image.
- Change uriScheme to type String.
- Change handledRPCs type to Integer. Change is for similar reasons that appServiceType should be sent as a String. Also handledRPCs as an integer will allow for better integration with policies.
- Defined functionIDs for new RPCs.
- Remove servicesSupported from AppServicesCapabilities. The available serviceTypes can be found in the appServices parameter. appServices -> updatedAppServiceRecord -> serviceManifest -> serviceType.
- Remove configuration for supportedAppServices in the ini file. Proposal states that core should be able to handle future unknown service types with conflicts with defining service types in core’s static ini file.
- Define policy table structure.

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

#675

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
  Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Jordyn Mackool

jacobkeeler:
A few potential additional revisions:
1. Remove the `defvalue` from `AppServiceActivation.setAsDefault`, since that would imply that the service is unset from the default if the parameter is omitted.
[...]

PR #679, opened 2019-02-13
0167-app-services (Accepted With Revisions)

JackLivio:
Make revisions as highlighted [here](https://github.com/smartdevicelink/sdl_evolution/issues/675)

```
The Steering Committee voted to accept this proposal with revisions. The author is to revert the change of renaming `AppServiceCapability` params in the `SystemCapability` struct to align with how the proposal was originally accepted.

* It was also noted that the only AppServiceTypes that will be included in the Q1 releases are `MEDIA`, `WEATHER` and `NAVIGATION`.
* ` The VOICE_ASSISTANT service data should be removed from the main proposal and put into a new section titled “Amendments” in the proposal with a note that it was removed from the first proposal.
* GENERIC, COMMUNICATION_VOIP, MESSAGING and TTS should be removed from this proposal as they will not be implemented in this iteration. When these specific AppServiceTypes are to be added, they will need to be submitted in separate proposals and voted on.
```


proposals/0167-app-services.md
```diff
@@ -978,44 +903,41 @@ The HMI will be responsible for the actual selection and activation of app servi
...
 
 <function name="AppServiceActivation" messagetype="request">
-  <param name="serviceId" type="String" mandatory="true" >
+  <param name="serviceID" type="String" mandatory="true" >
     <description>The ID of the service that should have an activation event take place on</description>
   </param>
-  <param name="serviceType" type="Common.ServiceType" mandatory="true">
-    <description>The service type the service should have the activation event occur on</description>
+  <param name="serviceType" type="String" mandatory="true">
+    <description>The type of service that is to be offered by this app. See AppServiceType for known enum equivalent types. Parameter is a string to allow for new service types to be used by apps on older versions of SDL Core.</description>
```
jacobkeeler: Should be removed, as was done in the response.

proposals/0167-app-services.md
```diff
@@ -978,44 +903,41 @@ The HMI will be responsible for the actual selection and activation of app servi
 
 ```xml
 <function name="GetAppServiceRecords" messagetype="request">
-  <param name="serviceTypes" type="Common.ServiceType" mandatory="false">
-    <description>If included, only service records of supplied type will be returned in response. If not included, all service records for all types will be returned.</description>
+  <param name="serviceType" type="String" mandatory="false">
+    <description>The type of service that is to be offered by this app. See AppServiceType for known enum equivalent types. Parameter is a string to allow for new service types to be used by apps on older versions of SDL Core.</description>
```
jacobkeeler: Description doesn't match for this RPC

JackLivio: Fixed

proposals/0167-app-services.md
JackLivio: Updated

PR #686, opened 2019-02-22
Support for Android custom RouterService

shiniwat:
This proposal is to improve the case where SDL application needs to work with custom RouterService. The definition of custom RouterService is varied, but this proposal refers to the case where a head unit requires an OEM-specific app that works as a RouterService. For instance, if an OEM uses specific Bluetooth server UUID that differs from SDL standard's one, that should be custom RouterService.

### DIFF
--- a/proposals/NNNN-support-for-android-custom-routerservice.md
+++ b/proposals/NNNN-support-for-android-custom-routerservice.md
@@ -84,7 +84,7 @@ The pseudo-code of new asynchronous methos in RouterServiceValidator looks like 
 
 		if(this.service == null){
 			// retrieveBestRouterServiceName works asynchronous, and calls FindConnectedRouterCallback when it finished to find
-			retrieveBestRouterServiceName(this.context, new FindConnectedRouterCallback() {
+			retrieveBestRouterServiceNameAsync(this.context, new FindConnectedRouterCallback() {
 				@Override
 				public void onFound(ComponentName component) {
 					service = component;
@@ -120,6 +120,66 @@ ValidationStatusCallback interface and FindConnectedRouterCallback looks as foll
 		void onFailed();
 	}
 
+```
+
+retrieveBestRouterServiceNameAsync uses AsyncTask, and look something like below:
+```java
+	private void retrieveBestRouterServiceNameAsync(Context context, FindConnectedRouterCallback callback) {
+		new FindRouterTask(callback).execute(context;
+	}
+```
+
+The pseudo-code of FindRoutertask will be:
+```java
+	class FindRouterTask extends AsyncTask<Context, Void, ComponentName> {
+		FindConnectedRouterCallback mCallback;
+
+		FindRouterTask(FindConnectedRouterCallback callback) {
+			mCallback = callback;
+		}
+
+		@Override
+		protected ComponentName doInBackground(final Context... contexts) {
+			if (RouterServiceValidator.this.service != null) {
+				return RouterServiceValidator.this.service;
+			}
+			List<SdlAppInfo> sdlAppInfoList = AndroidTools.querySdlAppInfo(contexts[0], new SdlAppInfo.BestRouterComparator());
+			if (sdlAppInfoList != null && !sdlAppInfoList.isEmpty()) {
+				SdlAppInfo lastItem = sdlAppInfoList.get(sdlAppInfoList.size()-1);
+				for (SdlAppInfo info: sdlAppInfoList) {
+					final boolean isLast = (info.equals(lastItem));
+					ComponentName name = info.getRouterServiceComponentName();
+					final SdlRouterStatusProvider provider = new SdlRouterStatusProvider(contexts[0], name, new SdlRouterStatusProvider.ConnectedStatusCallback() {
+						@Override
+						public void onConnectionStatusUpdate(boolean connected, ComponentName service, Context context) {
+							if (connected) {
+								if (mCallback != null) {
+									mCallback.onFound(service);
+								}
+							} else {
+								Log.d(TAG, "SdlRouterStatusProvider returns service=" + service + "; connected=" + connected);
+								if (isLast && mCallback != null) {
+									mCallback.onFailed();
+								}
+							}
+						}
+					});
+					provider.checkIsConnected();
+					provider.cancel();
+				}
+			}
+			return null;
+		}
+
+		@Override
+		protected void onPostExecute(ComponentName componentName) {
+			super.onPostExecute(componentName);
+			if (componentName != null && mCallback != null) {
+				mCallback.onFound(componentName);
+			}
+		}
+
+	}
 ```
 
 ### The caller of the asynchronous method
--- a/proposals/NNNN-support-for-android-custom-routerservice.md
+++ b/proposals/NNNN-support-for-android-custom-routerservice.md
@@ -41,16 +41,24 @@ Currently SdlProxy identify the RouterService in three steps:
 3. Before the app actually binds with the RouterService, it has validation logic, and the RouterService must be trusted RouterService. If the validation fails. then the app fails to find the trusted RouterService, and fallback to legacy Bluetooth mode.
 
 The problem is that step #2 and #3 unlikely find the custom RouterService, because custom RouterService is lowest order.
-Step #1 actually depends on the timing when the app calls queryForConnectedService. It is up to the application that when to calls queryForConnectedService, but [the integration-basic document](https://smartdevicelink.com/en/guides/android/getting-started/integration-basics/) suggests it should be called at Activitiy onCreate.
+Step #1 actually depends on the timing when the app calls queryForConnectedService. It is up to the application that when to calls queryForConnectedService, but [the integration-basic document](https://smartdevicelink.com/en/guides/android/getting-started/integration-basics/) suggests it should be called at Activitiy's onCreate.
 
 ## Proposed solution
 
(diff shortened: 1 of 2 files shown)


shiniwat:
Hi @jordynmackool 
I believe the proposal is ready for review. I hope you can proceed this proposal for sdl_evolution process. Thanks in advance.

proposals/NNNN-support-for-android-custom-routerservice.md
```diff
@@ -0,0 +1,275 @@
...
+                    }
+                }
+            } catch(PackageManager.NameNotFoundException e) {}
+        }
+    }
+```
+
+## Potential downsides
+
+- The proposed solution will use SdlRouterStatusProvider, which actually binds to other app's RouterService and asks if the RouterService has connected transports. This will be done in main thread, so the caller should execute it from worker thread. This can be done with AsyncTask anyway.
```
jordynmackool: Update to:
Should the `worker` thread be `working` thread?

- The proposed solution will use SdlRouterStatusProvider, which actually binds to other app's RouterService and asks if the RouterService has connected transports. This will be done in **the** main thread, so the caller should execute it from **the** worker thread. This can be done with AsyncTask anyway.

proposals/NNNN-support-for-android-custom-routerservice.md
```diff
@@ -0,0 +1,275 @@
...
+- Also when RouterService is bound from SdlRouterStatusProvider, RouterService should not enter to foreground. This is to avoid unneeded notification message coming up especially on Android 8 or above.
```
jordynmackool: Update to:
- Also when RouterService is bound from SdlRouterStatusProvider, RouterService should not enter to foreground. This is to avoid unneeded notification **messages** coming up especially on Android 8 or above.

proposals/NNNN-support-for-android-custom-routerservice.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Impact on existing code
+
+- RouterServiceValidator's validate will become asynchronous method.
+- The synchronous validate() method would be deprecated.
+
+## Alternatives considered
+- The solution would be beneficial for both custom RouterService and standard RouterService, because it actually increases the chance to find a RouterService that connects with head unit.
```
jordynmackool: Update to: 
- The solution would be beneficial for both custom RouterService and standard RouterService, because it actually increases the chance to find a RouterService that connects with **the** head unit.

proposals/NNNN-support-for-android-custom-routerservice.md
```diff
@@ -0,0 +1,275 @@
...
+But one thing to consider is that when finding the connected RouterService, I believe it is better to try the custom RouterService first, which is different order from what BestRouterComparator in SdlAppInfo offers. I think it is better to add another comparator for this purpose.
```
jordynmackool: Please re-word without using `I`

But one thing to consider is that when finding the connected RouterService, **It would be better to try the** custom RouterService first, which is different order from what BestRouterComparator in SdlAppInfo offers. **It is** **believed** to be better to add another comparator for this purpose.

proposals/NNNN-support-for-android-custom-routerservice.md
```diff
@@ -0,0 +1,275 @@
...
+        <intent-filter>
+            <action android:name="com.smartdevicelink.router.service" />
+        </intent-filter>
+        <meta-data android:name="sdl_custom_router" android:value="true" />
+        <meta-data android:name="sdl_router_version" android:value="@integer/sdl_router_service_version_value" />
+    </service>
+```
+
+When SDL application determines which RouterService to bind with, the app creates the list of SDL enabled apps, and checks to see if a RouterService whose transport connects with a head unit.
+If a custom RouterService is already connected with a head unit, and a SDL application can identify the RouterService correctly, the app should work fine. However, if not, the application instantiates "possibly the best" RouterService. In this case, custom RouterService is less likely chosen, because custom RouterService has lower priority than non-custom RouterService.
```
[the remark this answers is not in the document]
shiniwat: Thanks for the feedback. I have slightly modified the sentense to clarify the if-then logic in this sentense:

If a custom RouterService is already connected with a head unit, and if a SDL application can identify the RouterService correctly, then the app should work fine. However, if not, the application instantiates "possibly the best" RouterService. In this case, the custom RouterService is less likely chosen, because the custom RouterService has lower priority than the non-custom RouterService.

proposals/NNNN-support-for-android-custom-routerservice.md
```diff
@@ -0,0 +1,275 @@
...
+    -  Internally, SdlProxy has the priority order, the latest (which means having the newest version) non-custom RouteService has the priority.
+3. Before the app binds with the RouterService, it verifies that the RouterService is trusted one. If not, then the app falls back to legacy Bluetooth mode.
+
+The problem is that custom RouterService is unlikely chosen in step #2.
+Step #1 actually depends on the timing when an app calls queryForConnectedService. It is up to the application when queryForConnectedService is called, but [the integration-basic document](https://smartdevicelink.com/en/guides/android/getting-started/integration-basics/) suggests it should be called at Main Activitiy's onCreate.
+
+## Proposed solution
+
+Suppose a custom RouterService is basically designed for OEM-specific head unit, and the head unit works only with that specific custom RouterService.
+The current logic in previous section does not work very well in this scenario, because it heavily depends on the timing when the app calls SdlBroadcastReceiver#queryForConnectedService.
```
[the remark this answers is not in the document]
shiniwat: Yes, SdlBroadcastReceiver#queryForConnectedService is intentional.

proposals/NNNN-support-for-android-custom-routerservice.md
```diff
@@ -0,0 +1,275 @@
...
+
+2. Custom RouterService can not rely on other apps to be waken up, because of its priority order. Instead, it needs to start RouterService by himself.
+
+3. We can reuse the existing logic that verifies the trusted RouterService. If a RouterService is not trusted, SdlProxy can wake up the best possible RouterService in the same way as current proxy does. 
+
+In approach #1, the important point is the location where we find the RouterService its primary transport connects with head unit. I think it is in RouterServiceValidator class.
+
+### RouterServiceValidator should pay attention to currently connected RouterService.
+
+When would be the expected time to check if a RouterService connects with head unit or not? That would be right before the app binds to RouterService, and that should be done automatically without relying on user application.
```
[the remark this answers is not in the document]
shiniwat: Thanks for the suggestion. Updated as you suggested, but the sentense may look somewhat too long.

proposals/NNNN-support-for-android-custom-routerservice.md
shiniwat: "worker thread" is a kind of idiom, and intended.

shiniwat:
Hi @jordynmackool 
Many thanks for the feedback. I think most points you suggested are addressed already. Two are still open with my comments, but I have modified those two to clarify my intention. Any comments/suggestions are appreciated.

proposals/NNNN-support-for-android-custom-routerservice.md
```diff
@@ -25,42 +25,42 @@ Currently, AndroidManifest.xml can indicate the application has custom RouterSer
...
 
-In approach #1, the important point is the location where we find the RouterService its primary transport connects with head unit. I think it is in RouterServiceValidator class.
+In approach #1, the important point is the location where the RouterService's primary transport connects with the head unit. It should be the RouterServiceValidator class.
 
 ### RouterServiceValidator should pay attention to currently connected RouterService.
 
-When would be the expected time to check if a RouterService connects with head unit or not? That would be right before the app binds to RouterService, and that should be done automatically without relying on user application.
-The proposed solution is to add new asynchronous method to RouterServiceValidator, and calls it right before TransportManager connects to the RouterService.
+The expected time to check if a RouterService connects with head unit would be right before the app binds with the RouterService, and that should be done automatically without relying on user application.
```
jordynmackool: It does seem a little long, would this work:

The expected time to check if a RouterService connects with head unit would be right before the app binds with the **RouterService - this** should be done automatically without relying on user application.

proposals/NNNN-support-for-android-custom-routerservice.md
```diff
@@ -0,0 +1,275 @@
...
+2. If the step#1 fails, then SdlProxy wakes up the best possible RouterService.
+    -  Internally, SdlProxy has the priority order, the latest (which means having the newest version) non-custom RouteService has the priority.
+3. Before the app binds with the RouterService, it verifies that the RouterService is trusted one. If not, then the app falls back to legacy Bluetooth mode.
+
+The problem is that custom RouterService is unlikely chosen in step #2.
+Step #1 actually depends on the timing when an app calls queryForConnectedService. It is up to the application when queryForConnectedService is called, but [the integration-basic document](https://smartdevicelink.com/en/guides/android/getting-started/integration-basics/) suggests it should be called at Main Activitiy's onCreate.
+
+## Proposed solution
+
+Suppose a custom RouterService is basically designed for OEM-specific head unit, and the head unit works only with that specific custom RouterService.
```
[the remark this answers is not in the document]
jordynmackool: Please update

proposals/NNNN-support-for-android-custom-routerservice.md
shiniwat: Sounds good. Thanks.

shiniwat:
Hi @jordynmackool 
Thanks for the review. Updated the proposal in accordance with your feedback. I also made small update in code snippet. Hope this can move to next step. Thanks.

PR #705, opened 2019-04-12
Revise SDL 0220: support for Android custom RouterService

shiniwat:


shiniwat:
Hi @theresalech,
This is revised proposal of SDL 0220: support for Android custom RouterService. I have updated the proposal in accordance with steering committee held on Apr 9th. This is ready for review. Thanks in advance.

PR #711, opened 2019-04-23
Add iOS SDLManager RPC Subscriptions

joeljfischer:
This proposal improves subscribing to RPC requests, responses, and notifications on iOS and brings it more into alignment with the Java Suite library.


joeljfischer:

PR #719, opened 2019-05-03
SDL Server Version Tracking

nickschwab:
The goal of this proposal is enable the SDLC and Project Maintainer to track which SDL Server versions are in use by OEMs.


nickschwab:
@theresalech ready for review!

PR #725, opened 2019-05-09
Revise for SDL-0220: Support for Android custom RouterService

shiniwat:
(This is revised version of SDL-0220)


shiniwat:
Hi @theresalech 
This is revised version of SDL-0220. I made two major changes in this revision:
1) Code snippet has been updated based on the feedback from project maintainer.
2) bindToService modification and description about foregroundService have been removed.

This is ready for review.

shiniwat:
Hi @theresalech 
Thanks for the review! I have updated a code snippet slightly in order to take account of our internal review. Thanks.

PR #727, opened 2019-05-14
Add Pushing Buffer Support to AudioStreamManager

bilal-alsharifi:
This proposal is to add the ability to push an audio buffer to the `AudioStreamManager` so it supports playing audio from raw data in addition to supporting playing audio from actual files.


bilal-alsharifi:

Issue #691, opened 2019-03-06
[Accepted with Revisions] SDL 0220 - Support for Android custom RouterService


jordynmackool:
Hello SDL community,

The review of the revised "SDL 0220 - Support for Android custom RouterService" begins now and runs through May 21, 2019.  Previous reviews of "Support for Android custom RouterService" took place April 17 - May 7, 2019, March 27 - April 9, 2019, and March 6 - March 12, 2019. The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0220-support-for-android-custom-routerservice.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/691

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Jordyn Mackool

Program Manager - Livio
jordyn@livio.io

joeygrover:
I'm having a little trouble following this proposal. I believe there might be some misunderstanding of how the router service system works so hopefully I can clear them up while also providing a review.

>The definition of custom RouterService varies, [...]

It is only a designation given to an app that does not use the standard router service code base and must be flagged as such.

>When a SDL application determines which RouterService to bind with
>
An application only attempts to bind with an already connected router service. This normally happens when the router service sends an onConnected intent. The corner case happens when an app missed that intent and started by the user (the `queryForConnectedService` call)

>Step #1 actually depends on the timing of when the app calls queryForConnectedService.

Ok I think this is where things are misunderstood.

>Do not rely on the user app to call queryForConnectedService; instead, SdlProxy should automatically check to see if the RouterService connects with SDL Core whenever needed.

This is currently what the library does. Let me try to go step by step here

1. An app is installed and opened
2. App calls `queryForConnectedService`, no router service is running and bluetooth is off so no RS is started
[...]

PR #744, opened 2019-05-23
SDL 0220: accepted revision for SDL-0220 proposal

shiniwat:

### DIFF
--- a/proposals/0220-support-for-android-custom-routerservice.md
+++ b/proposals/0220-support-for-android-custom-routerservice.md
@@ -153,6 +153,7 @@ The pseudo-code of FindRouterTask will be:
 ```java
 	class FindRouterTask extends AsyncTask<Context, Void, ComponentName> {
 		FindConnectedRouterCallback mCallback;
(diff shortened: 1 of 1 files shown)


shiniwat:
Hi @theresalech,
This is accepted revision for SDL-0220 proposal based on the discussion at https://github.com/smartdevicelink/sdl_evolution/issues/691.

proposals/0220-support-for-android-custom-routerservice.md
```diff
@@ -245,35 +242,25 @@ validateAsync method is expected to be called in TransportManager, something lik
 	}
 
 	/**
-	 * start is now synonym of startValidate, and transport.start gets called in startTransport.  
+	 * start does validateAsync first, and then acxtually starts transport after the target RouterService is validated.
```
theresalech:
```suggestion
	 * start does validateAsync first, and then actually starts transport after the target RouterService is validated.
```

shiniwat: Oops, thanks for pointing that out!
I have updated the source branch to fix the typo.

PR #759, opened 2019-06-06
Update SDL-0234 with accepted revisions

joeljfischer:
* `rpc_spec` should be a submodule in each repository.
* Note that interfaces for the script should be identical across platforms.


joeljfischer:

Issue #730, opened 2019-05-15
[In Review] SDL 0230 - SPP resource management for Android


theresalech:
Hello SDL community,

The review of the revised proposal "SDL 0230 - SPP resource management for Android" begins now and runs through July 23, 2019.  Previous reviews of this proposal took place June 26 - July 9, 2019, June 5 - June 18, 2019, and May 15 - May 21, 2019.  The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0230-spp-resource-management-for-android.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/730

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Theresa Lech

Program Manager - Livio
theresa@livio.io

joeygrover:
| 1. There is no message defined as to what will be presented to the user. This should be defined here as to what is an acceptable message to the developer while taking into consideration they might have already started driving by the time they see this message.

If each app developer is forced to decide what should go into that notification, it adds the burden onto them to decide what it should say and the SDLC will not have control over such messaging. This can lead to confusion to the user when inconsistent messaging is present.

| 2. Legacy Mode is required for apps to work on IVI systems that use v1 of the SDL Protocol. If a developer selects to not allow legacy mode they will not work on those systems and it should be understood that is a risk.

| 3. Legacy mode is also an important part of ensuring apps are able to display on an IVI system even if the router service feature fails. Allowing apps to prevent this will likely lead to confusion on the users side that they do not understand why some apps show up and some apps do not. This is an undesirable user experience.

| 4. Disallowing legacy mode by default does not seem acceptable because of point 2 and 3. It should default to true if implemented.

[...]

PR #781, opened 2019-07-16
Another revision SDL-0230: SPP resource management

shiniwat:
RFCOMM is a connection-oriented, streaming transport over Bluetooth, which is known as Serial Port Profile (SPP) in Android. 
SDL Proxy uses BluetoothServerSocket to listen on a SPP channel, and it is used for primary transport.
On an Android device, the number of SPP resources is limited, and BluetoothServerSocket fails to accept connections when SPP channel runs out of available resources.
Current SDL Proxy does not handle that case very well. This proposal is to improve the SPP resource management in SDL Android.


shiniwat:
Hi @theresalech,
This is revised SDL-0230 proposal based on the discussion at https://github.com/smartdevicelink/sdl_evolution/issues/730.

proposals/0230-spp-resource-management-for-android.md
```diff
@@ -22,15 +22,15 @@ This proposal focuses on the case where Proxy detects the SPP resource error, an
 
 ### Detect the case where BluetoothServerSocket fails to accept a connection from head unit. 
 Even though we could detect the case, we cannot increase the number of available SPP resources, because they are used by other apps.
-All we can do in this case is notify users that SPP channel runs out of available resources, and let users close some apps that may use the BluetoothSocket. It's not practical to show SPP service records that are used by Bluetooth adapter. The proposed approach is to notify users that we're running out of resources, and gives more information when the user requests it.
+All we can do in this case is notify users that SPP channel runs out of available resources, and let users close some apps that may use the BluetoothSocket. It's not practical to show SPP service records that are used by Bluetooth adapter. The proposed approach is to notify users that we're running out of resources, and gives more information if needed (though this proposal does not include giving more information part).
```
theresalech:
```suggestion
All we can do in this case is notify users that SPP channel runs out of available resources, and let users close some apps that may use the BluetoothSocket. It's not practical to show SPP service records that are used by Bluetooth adapter. The proposed approach is to notify users that we're running out of resources, and gives more information if needed (though this proposal does not include the giving more information part).
```

proposals/0230-spp-resource-management-for-android.md
```diff
@@ -171,49 +191,21 @@ When user opens "manage notifications", the app's notification setting will be s
...
     <string name="spp_out_of_resource">Too many apps are using Bluetooth</string>
     <string name="notification_title">SmartDeviceLink</string>
-    <string name="spp_out_of_resource_message">There are too many bluetooth apps running on your device. Please close some of them and try to re-connect</string>
-    <string name="spp_out_of_resource_possible_apps">Following apps may use Bluetooth: </string>
-    <string name="button_ok">OK</string>
     <string name="sdl_error_notification_channel_name">SDL Error</string>
 ```
 
-The strings above are TBD (requires review).
+The strings above needs to be reviewed by steering committee.
```
theresalech:
```suggestion
The strings above need to be reviewed by SDLC steering committee.
```

proposals/0230-spp-resource-management-for-android.md
```diff
@@ -228,68 +220,10 @@ Regarding who is responsible for the localization process, two steps are propose
...
-	}
-```
-
-Showing the list to the user will be optional (at least developers can choose), because we may need to determine if it is worth doing or not.
-
 ## Potential downsides
 
 In Fig. 3 (Manage notification category setting), if user changed the behavior to "No sound", user won't notice anything even if SPP error occurred. This seems to be a potential issue, but it is up to the user's configuration.
-
+On the other hand, we should be very aware of the fact that SDL proxy notifies error with sound and/or vibration when SPP error occurred.
```
theresalech:
```suggestion
On the other hand, we should be very aware of the fact that SDL proxy notifies error with sound and/or vibration when SPP error occurs.
```

shiniwat:
Hi @theresalech,
Thanks for the feedback! I have updated the proposal as you suggested.

PR #793, opened 2019-07-25
Accepted with revision: SDL-0230

shiniwat:
RFCOMM is a connection-oriented, streaming transport over Bluetooth, which is known as Serial Port Profile (SPP) in Android. 
SDL Proxy uses BluetoothServerSocket to listen on a SPP channel, and it is used for primary transport.
On an Android device, the number of SPP resources is limited, and BluetoothServerSocket fails to accept connections when SPP channel runs out of available resources.
Current SDL Proxy does not handle that case very well. This proposal is to improve the SPP resource management in SDL Android.


shiniwat:
Hi @theresalech 
This is the PR for accepted revision, which reflects the discussion at https://github.com/smartdevicelink/sdl_evolution/issues/730.

PR #803, opened 2019-08-01
Add App Services to HMICapabilities

joeygrover:
[SDL-0167](0167-app-services.md) introduced the App Services functionality. It included a new capability, however that capability was not included into the `HMICapabilities` struct as all other system capabilities have been.


joeygrover:
@theresalech ready for review

PR #805, opened 2019-08-06
Revise SDL-0179 Pixel Density and Scale

NicoleYarroch:
# Introduction
Update SDL-0179 Pixel Density and Scale by adding a min and max value for the `scale` parameter in the `VideoStreamingCapability`.

# Motivation
The current implementation allows for zero and negative `scale` values. A min and max value should be added to prevent invalid `scale` values.

# Proposed Solution
The proposed changes are as follows:
```
<param name="scale" type="Float" minvalue="1" maxvalue="10" mandatory="false">
```
The `minvalue` and `maxvalue` were added.

# Potential downsides
No downsides identified; this is a very small update.

# Impact on existing code
The implementations will have to be updated

# Alternatives considered
No alternatives considered


NicoleYarroch:
@theresalech Ready for review

PR #817, opened 2019-09-03
Revisions - SDL 0248 - HMI Policy Table Update using vehicle modem

kshala-ford:
This proposal describes an enhanced policy table update PTU sequence where the IVI system sends the policy table snapshots (PTS) directly to the policy server using the in-vehicle modem and returns a policy table update (PTU) back to SDL Core without using any `SystemRequest` RPC.

This PR provides revisions as per https://github.com/smartdevicelink/sdl_evolution/issues/813#issuecomment-525801811


kshala-ford:
@theresalech this PR is ready for review

PR #908, opened 2020-01-13
Screen Manager Subscribe Buttons

joeljfischer:
This proposal introduces a screen manager feature for developers to subscribe to hard buttons instead of sending subscribe / unsubscribe RPCs.


joeljfischer:
@theresalech This proposal is ready for review

PR #911, opened 2020-01-15
SDL 0259 - Disabled Softbuttons

MichaelCrimando:
This feature would allow apps to send custom softbuttons to display on screen but they would appear disabled. This also allows an app to subscribe to and disable SDL predefined buttons.

These are the final changes for the "Accepted with revisions" from the SDLC

### DIFF
--- a/proposals/0259-DisabledSoftbuttons.md
+++ b/proposals/0259-DisabledSoftbuttons.md
@@ -1,5 +1,6 @@
 
 
+
 # Disabled Softbuttons
 
 * Proposal: [SDL-0259](0259-DisabledSoftbuttons.md)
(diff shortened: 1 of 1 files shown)


MichaelCrimando:
@joeljfischer @joeygrover @kshala-ford 
Can one of you please give this a final review and make sure I got everything?

joeljfischer:
It appears that there was confusion in the comments about Joey's original comment #4, which was that the RPC parameter should be changed from `isDisabled` to `disabled`. I don't ultimately know if this was agreed to in the final SDLC explanation. I don't see it in Theresa's notes. @theresalech @joeygrover

joeygrover:
@joeljfischer Yea that was part of it. It's not a deal breaker for me; I just thought it was a descriptive enough name for a boolean param to be `disabled` instead of adding the modifier to it. Like you said, I am unsure if it was actually agreed upon or not in the call. So  since it doesn't appear in the rationale of the vote we can either leave it as is or we can ask the SDLC in the next meeting if the change was accepted. I'm good either way.

MichaelCrimando:
@joeljfischer  and @joeygrover 
Ok updated the proposal as you were saying, does that look like your intent?

It's not really a big change in understanding, I cant see why anyone would dissent to that.

joeljfischer:
@MichaelCrimando I understand and agree that it's a small change that's highly unlikely to have controversy, but we still have to bring it to the steering committee for process reasons.

MichaelCrimando:
@joeljfischer ok thats fine, are the changes I put in good to go?

joeljfischer:
I think it's ready to bring it to the SDLC, yeah

theresalech:
@MichaelCrimando @joeljfischer I've included this on the agenda for this week's meeting, thanks!

theresalech:
The Steering Committee confirmed agreement with changing the RPC parameter from `isDisabled` to `disabled`.  Merging!

PR #925, opened 2020-01-31
SDL-0274 Add preferred FPS to VideoStreamingCapability (accepted revision)

shiniwat:
This proposal extends SDL video streaming feature by adding preferred FPS to `VideoStreamingCapability`, so that frame rate of video streaming can be adjusted per OEM.

### DIFF
--- a/proposals/0274-add-preferred-FPS.md
+++ b/proposals/0274-add-preferred-FPS.md
@@ -3,7 +3,7 @@
 * Proposal: [SDL-0274](0274-add-preferred-FPS.md)
 * Author: [Shinichi Watanabe](https://github.com/shiniwat)
 * Status: **Accepted with Revisions**
-* Impacted Platforms: [Core / iOS / Java Suite / RPC]
+* Impacted Platforms: [Core / iOS / Java Suite / JavaScript Suite / RPC]
 
 ## Introduction
 
@@ -32,7 +32,7 @@ Add `preferredFPS` to `VideoStreamingCapability` struct in both APIs.
         ...
         <!-- new param -->
         <param name="preferredFPS" type="Integer" minvalue="0" maxvalue="2147483647" mandatory="false">
-            <description>Preferred frame rate per second. Mobile application should take this value into account for capturing and encoding video frame.</description>
+            <description>Preferred frame rate per second. Mobile application should take this value into account for capturing and encoding video frame, but mobile application should also take care of the case where mobile device's hardware performance is constrained. It is up to mobile application to determine the actual frame rate.</description>
         </param>
     </struct>
 ```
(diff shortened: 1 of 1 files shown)


shiniwat:
Hi @theresalech, this PR is to align for accepted revision, and ready for review. Thanks.

proposals/0274-add-preferred-FPS.md
```diff
@@ -82,7 +82,22 @@ static NSUInteger const defaultFrameRate = 15;
 
 }
 ```
-Note that `videoEncoderSettings` property must be `NSMutableDictionary` as it now would be mutable.
+
+In `SDLStreamingVideoLifecycleManager#didEnterStateVideoStreamStarting`, current `videoEncoderSettings` are overwritten by `customEncoderSettings`.
```
theresalech:
```suggestion
In `SDLStreamingVideoLifecycleManager.didEnterStateVideoStreamStarting`, current `videoEncoderSettings` are overwritten by `customEncoderSettings`.

```

proposals/0274-add-preferred-FPS.md
```diff
@@ -82,7 +82,22 @@ static NSUInteger const defaultFrameRate = 15;
...
+            if ([(NSString *)key isEqualToString:(__bridge NSString *)kVTCompressionPropertyKey_ExpectedFrameRate] == YES) {
+                // do NOT override frame rate if custom setting's frame rate is higher than current setting's one.
+                if ([self.customEncoderSettings valueForKey:key] < self.videoEncoderSettings[key]) {
+                    self.videoEncoderSettings[key] = [self.customEncoderSettings valueForKey:key];
+                }
+            } else {
+                self.videoEncoderSettings[key] = [self.customEncoderSettings valueForKey:key];
+            }
+        }
+```
```
theresalech: This should take into account the new `preferredFPS` key.

proposals/0274-add-preferred-FPS.md
```diff
@@ -32,7 +32,7 @@ Add `preferredFPS` to `VideoStreamingCapability` struct in both APIs.
         ...
         <!-- new param -->
         <param name="preferredFPS" type="Integer" minvalue="0" maxvalue="2147483647" mandatory="false">
-            <description>Preferred frame rate per second. Mobile application should take this value into account for capturing and encoding video frame.</description>
+            <description>Preferred frame rate per second. Mobile application should take this value into account for capturing and encoding video frame, but mobile application should also take care of the case where mobile device's hardware performance is constrained. It is up to mobile application to determine the actual frame rate.</description>
```
theresalech:
```suggestion
            <description>Preferred frame rate per second. Mobile application should take this value into account for capturing and encoding video frame, but the app library should also take care of the case where the mobile device's hardware performance is constrained. It is up to the app library to determine the actual frame rate.</description>

```

shiniwat: Well, the suggested change does not correct. If iOS SDL application that uses SDLCarWindow and LifecycleManager, sdl_ios library will take care of the frame rate. In other words, frame rate is still up to mobile application if some iOS apps that do not use SDLCarWindow and/or Android application running with current sdl_java_suite library. I have no good idea how to explain the situation appropriately, however.

shiniwat:
Hi @theresalech, thanks for the feedback. I have updated the PR that hopefully resolve the first and second suggestion. Before the change, the proposal did not make it clear that preferredFPS value is stored in videoEncoderSettings already in earlier part of the code snippet.
The third one (description in MOBILE_API) still needs some discussion and your feedback is highly appreciated. Thanks!

proposals/0274-add-preferred-FPS.md
joeljfischer: What about this instead?

`The preferred frame rate per second of the head unit. The mobile application / app library may take other factors into account that constrain the frame rate lower than this value, but it should not streaming at a higher FPS than this value.`

shiniwat: @joeljfischer, thanks for the suggestion. Sounds good to me.

shiniwat:
Hi @theresalech, updated the description accordingly. Thanks!

PR #951, opened 2020-02-27
Revise SDL-0234 - Proxy Library RPC Generation proposal 

bilal-alsharifi:
## Introduction
Update SDL-0234 Proxy Library RPC Generation proposal to not implement custom mappings in JavaSuite and iOS.

## Motivation
Implementing the proposal with custom mappings for every corner case will make the code hard to maintain the future. It will be much cleaner if the generator scripts generate the code directly from the RPC spec with no case by case handling.

## Proposed solution
Update the proposal to not implement custom mappings in JavaSuite and iOS, as creating corner case mappings was found to be a larger maintenance burden than maintaining the existing classes themselves.  Custom mappings can still occur in general cases that can be applied to everything of a particular type, e.g. all enum names that start with a number need to be prefixed by `_`. But adding  corner cases for specific classes should be avoided.

## Potential downsides
The code that is generated for existing classes may be a bit different than the existing code. But that should be fine as the generator scripts will only be used for new RPCs and not the existing ones.

## Impact on existing code
The mobile libraries will need to be updated.

## Alternatives considered
Implement the mapping for all corner cases. That will make the scripts very complex and hard to maintain in the future.

### DIFF
--- a/proposals/0234-proxy-rpc-generation.md
+++ b/proposals/0234-proxy-rpc-generation.md
@@ -45,6 +45,7 @@ The script should take a `MOBILE_API` XML spec as input and a location to output
     The submodule reference was updated with more recent parser scripts but the generator scripts are not updated
 
 13. Python 2 support ends in 2020. Therfore, all python scripts should be developed for Python 3. This includes scripts from the InterfaceBuilder that will be refactored.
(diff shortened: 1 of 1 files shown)


bilal-alsharifi:
@theresalech The PR is ready for review.

PR #976, opened 2020-03-19
Accepted revisions for SDL 0290 - Modify The Range of APT's maxDuration

zhouxin627:
**Updates**

- Clarify that the RPC Spec and Core are both impacted platforms which will require major version changes upon implementation of this proposal.


zhouxin627:
@theresalech Hi, this is ready for review.

PR #978, opened 2020-03-19
Accpeted Revisions for SDL 0283 ScreenManager Slider and Scrollable Message

joeljfischer:
## Revisions
* Remove SDLScreenManagerDefaultTimeout
* SDLScreenManagerSliderCompletionHandler selectedPosition should be an integer and be -1 if there's an error


joeljfischer:
@theresalech this is ready for review

PR #980, opened 2020-03-24
WiFi: Sharing SSID and password with Mobile Proxy

LuxoftAKutsan:
In the current implementation of SDL over WiFi transport, IP address and port are shared with Mobile proxy. However, the prerequisite for using SDL over WiFi transport is that the mobile device should be connected to WiFi. In some cases, entering a password for the access point created by HMI could be tedious and it would be a lot more convenient for users if devices could automatically connect via WiFi.

If WiFi credentials are automatically generated by the IVI system, the password could contain special characters or it could be just too long. It would be inconvenient for the user to enter such passwords in the HMI. Another possible scenario is, the HMI could keep changing passwords to improve the security of the network. In such cases, a user might need to connect to the WiFi network every time the password is changed.


LuxoftAKutsan:
Hi @theresalech please add this proposal changes to SDLC agenda.

theresalech:
Hi @LuxoftAKutsan - the agenda for tonight's meeting has already been set and sent to attendees, but if you have something regarding this proposal (outside of the typical SDL Evolution Process) that you'd like to discuss, you can bring it as a walk-in if there's time. Otherwise, we will be following the standard process and I'll review this proposal for formatting/readability and mark as ready for review once my feedback has been addressed. Thanks!

LuxoftAKutsan:
Changes moved to Alternative considered section of the proposal in PR :  https://github.com/smartdevicelink/sdl_evolution/pull/981

PR #985, opened 2020-03-26
Requested Revisions for SDL 0294 - App Service Messaging

joeljfischer:
### Updates
* Make `MessagingServiceData.recentMessageThreads` an array
* Update `MessageContact` with documentation and add a `contactID` parameter for `SendMessage` RPC
* Add the `SendMessage` RPC to allow for service consumers to send messages through a provider, if permitted
* Add `PerformAppServiceInteraction.rpcRequest` / `rpcResponse` to allow for sending RPCs to inactive services
* Update `GetAppServiceData` with a `serviceID` parameter to allow for subscribing to the data of inactive services; this is especially important for the messaging service since effectively all services should be "active"
* Update downsides with (2) for privacy concerns around consumers having access to message text
* Update downsides with (3) for concerns around bad actor apps sending spam messages through messaging service providers
* Update "Impact on Existing Code" to note that the developer portal / default policies should be updated to restrict `SendMessage` as well.
* Updated alternative (1) to note that it is incorportated
* Added alternative (2) with another possible solution to subscribing to inactive services
* Added HMI_API details

### DIFF
--- a/proposals/0294-app-service-messaging.md
+++ b/proposals/0294-app-service-messaging.md
@@ -8,9 +8,10 @@
 This proposal adds a new "messaging" app service for apps that provide messaging features.
 
 ## Motivation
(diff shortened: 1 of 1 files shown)


joeljfischer:
@theresalech This is ready for review, thank you!

PR #986, opened 2020-03-26
Revise SDL-0234 - Proxy Library RPC Generation proposal - Keyword Avoidance

joeygrover:
## Introduction
Update SDL-0234 Proxy Library RPC Generation proposal to avoid keywords that already exist in base classes which would conflict.

## Motivation
Implementing the proposal without taking into account the already existing method signatures has created collisions in the RPC classes. For example, the `CancelInteraction` RPC request has a parameter `functionID`, based on current rules the generator scripts would create getters and setters as `getFunctionID()` and `setFunctionID(..)`. However, because the `RPCMessage.java` class already has a `getFunctionID()`, this method would effectively be overridden or the code would not compile as return types could differ. Then calling the incorrect method would only return the function ID associated with the parameter `CancelInteraction`. 

## Proposed solution

Update the proposal to include a rule that takes into account a set of keywords that currently exist in any of the three client side libraries (iOS, Java Suite, and JavaScript suite), and avoid creating method signatures that collide. 

Keywords will be stored in a flat text file in the RPC spec repo. The scripts can access this file already based on the spec repo being a submodule. Keywords will be newline separated. The script should load in these keywords to ensure they are avoided in all three client libraries.

If a collision is detected from a keyword, all internal references of that keyword should be appended with `Param`.

For example, `CancelInteraction`

```java

public class CancelInteraction extends RpcRequest{

    private static final KEY_FUNCTION_ID_PARAM = "functionID"; //This actual string value must remain the same as the rpc spec
    FunctionID getFunctionID(){
        return functionID; //This would be equal to FunctionID.CANCEL_INTERACTION;
    }
    
    Integer getFunctionIDParam(){
    	return getInteger( KEY_FUNCTION_ID_PARAM);
    }
    
    void setFunctionIDParam(Integer functionIDParam){
        setParameters(KEY_FUNCTION_ID_PARAM,functionIDParam);
    }
}

```

### Unifying APIs between libraries

The addition of a keyword avoidance feature into the generator scripts with clear and concurrent logic the library APIs can be unified. This will present a much more intuitive API to developers who switch between using the libraries. Otherwise, the likelihood of creating parity between methods is very high. 

For example any of the following methods could be used as naming conventions for the method signature of the getter for `functionID` param.

```java
getFunctionId()
getFunctionid()
getfunctionid()
getfunctionID()
get_FunctionID()
getFunctionID_()
getParamFunctionID()
getFunctionIDParam()
getInteractionFunctionID()
getCancelInteractionFunctionID()

```
It can obviously be almost infinite of possible naming conventions used to avoid the collision. Creating a standard methodology is imperative for creating well designed APIs.

### Core's Generator Script

Since core is the only project currently on the "server" side it does not need to use this keyword file to generate RPC methods.

## Potential downsides
This will slightly increase the scope of the generator scripts, however we avoid making breaking changes to the libraries in the future by including this change now.

## Impact on existing code
The mobile and web libraries will need to be updated.

## Alternatives considered
- Ignoring these cases and handling them manually. This adds extra burden to the creation of new RPCs where this feature is to be used the most. It also introduces the likelihood that the APIs will differ between each project.


joeygrover:
@theresalech ready for review!

proposals/0234-proxy-rpc-generation.md
```diff
@@ -44,8 +44,9 @@ The script should take a `MOBILE_API` XML spec as input and a location to output
...
+13. Python 2 support ends in 2020. Therefore, all python scripts should be developed for Python 3. This includes scripts from the InterfaceBuilder that will be refactored.
 14. The generator script should only be used for newly created RPCs. Existing RPC classes should not be overwritten by generated code because the existing code may be slightly different than the generated one.
+15. The generator scripts should take in a set of keywords that must be avoided while creating RPC methods. These keywords will be stored a flat text file separated by newlines. If a parameter or its potential method signatures (get/set/etc) collides with a keyword, the parameter, for internal reference only, should be appended with `Param`. For example, the param of `functionID` exists in the `CancelInteraction` RPC, but the `getFunctionID()` method signatures already exist in base classes and would be included in the keyword file. Therefore, the generator should create methods for this param as `getFunctionIDParam()` and `setFunctionIDParam(...)`.
```
theresalech:
```suggestion
15. The generator scripts should take in a set of keywords that must be avoided while creating RPC methods. These keywords will be stored in a flat text file separated by newlines. If a parameter or its potential method signatures (get/set/etc) collides with a keyword, the parameter, for internal reference only, should be appended with `Param`. For example, the param of `functionID` exists in the `CancelInteraction` RPC, but the `getFunctionID()` method signatures already exist in base classes and would be included in the keyword file. Therefore, the generator should create methods for this param as `getFunctionIDParam()` and `setFunctionIDParam(...)`.
```

joeygrover:
@theresalech updated and ready for re-review

PR #998, opened 2020-04-21
SDL Device Listener

joeygrover:
The SDL Android library relies on a complex multiplexing system that operates through a foreground service. This service is started based on the connection of the Media and Phone bluetooth profiles. Since SDL is set up in a way that the mobile device acts as the listening server of a bluetooth SPP connection, the router service must be started and waits for incoming connections without knowing if anything will connect to it. This leads to the service having to start in the foreground for all bluetooth connections. One possible way to avoid this is introducing a new prerequisite before starting the router service. This is done by adding a new class called `SDLDeviceListener` that will be activated in the `SDLBroadcastReceiver` to decide if the router service should be started or not.


joeygrover:
@theresalech ready for review

proposals/NNNN-SDL-device-listener.md
```diff
@@ -0,0 +1,244 @@
...
+
+The SDL library already attempts to prevent the user from seeing the notification. This is done by quickly enters the foreground as required by the Android OS, then within 10 milliseconds, the service exits the foreground, notification removed, and notification channel deleted. However, based on testing different phones it was obvious that this has a multitude of different results. Some phones do exactly as expected and the notification is cleared before presented to the user. Other devices and OS versions will leave the notification in the tray for roughly 5 seconds until it is removed. The project maintainers have refactored the logic to avoid this for the majority of device and OS combinations, but some devices do not honor the APIs correctly.
```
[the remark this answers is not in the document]
theresalech: >This is done by quickly enters the foreground as required by the Android OS, then within 10 milliseconds, the service exits the foreground, notification removed, and notification channel deleted." 

Please update to "This is done by quickly entering the foreground as required by the Android OS, then within 10 milliseconds, the service exits the foreground, notification is removed, and notification channel is deleted."

proposals/NNNN-SDL-device-listener.md
```diff
@@ -0,0 +1,244 @@
...
+
+Because of these adverse situations, the library should alter its logic that handles the starting of the `SdlRouterService`. This can be accomplished by creating a new gatekeeper class that prevents the service from ever being started until there has been a confirmed SDL connection via the standard SDL UUID. This logic can be inserted into the `SdlBroadcastReceiver` without a lot of overhead to existing code.
+
+## Proposed solution
+
+The solution will be to create a temporary bluetooth server socket in the broadcast receiver that receives the bluetooth connection intents. A new thread will be spawned off that will handle the listening. If a connection is made with the device, the bluetooth connection will be closed, and then the `SdlRouterService` will be started up directly to the foreground. A timeout handler will be present to keep the socket alive for a short duration without allowing it to run forever. Once the timeout expires, if there was no connection made, the socket will stop listening and the router service will not be started. In either case the MAC address will be saved into preferences with the results of a successful connection or not.
```
theresalech, line 23:
```suggestion
The solution will be to create a temporary bluetooth server socket in the broadcast receiver that receives the bluetooth connection intents. A new thread will be spawned off that which will handle the listening. If a connection is made with the device, the bluetooth connection will be closed, and then the `SdlRouterService` will be started up directly to the foreground. A timeout handler will be present to keep the socket alive for a short duration without allowing it to run forever. Once the timeout expires, if there was no connection made, the socket will stop listening and the router service will not be started. In either case the MAC address will be saved into preferences with the results of a successful connection or not. 
```

theresalech: >A new thread will be spawned off that will handle the listening.
Please update to "A new thread will be spawned off that which will handle the listening."

proposals/NNNN-SDL-device-listener.md
```diff
@@ -0,0 +1,244 @@
...
+
+On the next connection with that same device, via the MAC address, we can avoid this step if the device has been recognized as an SDL enabled device. If the device has not made a connection, the timeout window can be shortened. 
+
+Leaving a small window of time to allow for connections at this stage is very important. The first time a device is connected, it might be plausible the SDL connection was just too slow and the following connection might catch the right timeout. The goal is make sure there isn't a missed SDL enabled device while taking into consideration that most users do not interact with such devices.
```
theresalech, line 27:
```suggestion
Leaving a small window of time to allow for connections at this stage is very important. The first time a device is connected, it might be plausible the SDL connection was just too slow and the following connection might catch the right timeout. The goal is to make sure there isn't a missed SDL enabled device while taking into consideration that most users do not interact with such devices.
```

proposals/NNNN-SDL-device-listener.md
```diff
@@ -0,0 +1,244 @@
...
+
+<img src="../assets/proposals/NNNN-SDL-device-listener/flow_chart.png" alt="POI app service example" class="inline" height= "75%" width= "75%" /> 
+
+### SDLDeviceListener
+
+This class will be a small, scoped class that will be instantiated into the `SdlBroadcastReceiver` code base. See the Appendix for the sample code.
+
+#### Which apps create the SDLDeviceListener?
+
+One of the main reasons the `SdlRouterService` functionality was created was to avoid the case where an Android device didn't have enough RFCOMM channels to support each app hosting their own. Obviously this proposal doesn't want to run into that same situation. Therefore, each app will use the same logic it does now to start an `SdlRouterService`, but instead of each app starting the service, only the app that has the service to be started will be the one to instantiate an `SDLDeviceListener ` instance. All other apps will simply return out of that logic block. This will keep the RFCOMM channels used to only 1 per BT connection. There is the caveat that if any older apps are installed, the old logic must be followed, so this feature will be dynamically enabled over time.
```
theresalech, line 40:
```suggestion
One of the main reasons the `SdlRouterService` functionality was created was to avoid the case where an Android device didn't have enough RFCOMM channels to support each app hosting their own. Obviously this proposal doesn't want to run into that same situation. Therefore, each app will use the same logic it does now to start an `SdlRouterService`, but instead of each app starting the service, only the app that has the service to be started will be the one to instantiate an `SDLDeviceListener` instance. All other apps will simply return out of that logic block. This will keep the RFCOMM channels used to only 1 per bluetooth connection. There is the caveat that if any older apps are installed, the old logic must be followed, so this feature will be dynamically enabled over time.
```

theresalech: Please remove extra space after "SDLDeviceListener" and before the backtick.  Also, change "BT" to "bluetooth"

joeygrover:
@theresalech updated and ready for review again

PR #1009, opened 2020-05-12
Add a Reason Parameter to All Protocol NAKs

joeljfischer:
This proposal seeks to add a "reason" string parameter to all NAKs in the protocol spec.


joeljfischer:
@theresalech this proposal is ready for review

PR #1015, opened 2020-05-20
Revise SDL-0278 - Screen Manager Template Management

NicoleYarroch:
# Introduction
Update SDL-0278 Screen Manager Template Management by adding new APIs that allow the update of color schemes, text, graphics, buttons, and template title at the same time the template is being updated.

# Motivation
The current implementation only allows a simple template change. Since the `SetDisplayLayout` and `Show` requests allow for simultaneous update of the color schemes and template UI, new API was added to update those features as well. 

# Proposed Solution
The proposed public API was updated from:
```
- (void)changeLayout:(NSString *)newLayout withCompletionHandler:(SDLScreenManagerUpdateCompletionHandler)handler;
```
to the following:
```
- (void)changeLayout:(SDLTemplateConfiguration *)templateConfiguration withCompletionHandler:(nullable SDLScreenManagerUpdateCompletionHandler)handler;
```
1. The method now takes a `SDLTemplateConfiguration` (this struct already exists) instead of a `String` so the template and color schemes can be updated at the same time. 
1. The `SDLScreenManagerUpdateCompletionHandler` can now be set to `nil` if the developer does not care whether or not the template update succeeds. 

## Implementation Update Details
If the template update is batched with text, graphic updates on v6.0+ then all updates will be sent with one `Show` RPC (if buttons are also batched in the update, it will currently require a another `Show` request - this implementation detail could change in the future). On systems <v6.0, the `Show` request will be sent after  the `SetDisplayLayout`. If the layout update fails while batching, then the text, graphics, buttons or template title will also not be updated.

1. When connected to systems running RPC v6.0+, the `ScreenManager` will send `Show` request to change the template. If the template update is batched with other updates, the  `ScreenManager` will upload the graphics before sending a single `Show` request with the template, text, and graphic information. The current implementation of the sub-managers will require the soft buttons to be sent after the rest of the template update in a separate `Show` request, but that implementation detail could change in the future. 
 1. When connected to systems running RPC <v6.0, the screen manager will send a `SetDisplayLayout` to change the template. If the template update is batched with other updates, on receipt of a successful `SetDisplayLayout` response, the graphics will first be uploaded and then a `Show` request will be sent by the `ScreenManager` to update the text, graphic, and button information. If the `SetDisplayLayout` request fails, then the graphics and the `Show` request will not be sent in order to make sure that the behavior of the screen manager on RPC <v6.0 is consistent with RPC v6.0+. The current implementation of the sub-managers will require the soft buttons to be sent in a separate `Show` request after the rest of the template update (assuming the template update succeeds), but that implementation detail could change in the future.
 1. If the system is running RPC v5.0+ but less than v6.0, the template's day and night color schemes will be updated via the  `SetDisplayLayout` request. If the system is running RPC v6.0+, the template's day and night color schemes will be updated via the `Show` request's `templateConfiguration`. 

# Potential downsides
This update will need to wait until the existing Text and Graphic manager is updated to use queues so the Text and Graphic Manager and Soft Button Manager can be used sequentially to send the `Show` requests as needed. 

# Impact on existing code
The implementations will have to be updated.

# Alternatives considered
No alternatives considered.


proposals/0278-screenmanager-layout-management.md
```diff
@@ -11,33 +11,78 @@ Adds a screen manager method for changing the current layout to a new layout.
...
 The manager layer should be able to handle most tasks that used to be handled by RPCs, and the Screen Manager in particular should be able to handle manipulating the screen and templates without requiring the developer to fall back to RPCs in any cases.
 
-One big area that it is currently missing is the ability to manage changing templates, currently done with the `SetDisplayLayout` RPC (on RPC spec <6.0) and `Show` RPC (on RPC spec >6.0). This proposal adds a way to change the layout while taking care of backward compatibility concerns.
+One big area that it is currently missing is the ability to manage changing templates, currently done with the `SetDisplayLayout` RPC (on RPC spec <6.0) and `Show` RPC (on RPC spec >=6.0). This proposal adds a way to change the layout while taking care of backward compatibility concerns.
 
 ## Proposed solution
-We will add a new screen manager method for changing the current layout to a new layout.
+We will add a new screen manager method for changing the current layout to a new layout. This also includes updating the template's night and day color schemes when the feature is available on systems supporting RPC spec 5.0.0 or newer.
```
joeljfischer, line 17:
```suggestion
We will add a new screen manager method for changing the current layout to a new layout. This also includes updating the template's data immediately on v6.0+ and sequentially on <v6.0, as well as the template's color schemes on RPC v5.0+ systems.
```

proposals/0278-screenmanager-layout-management.md
```diff
@@ -11,33 +11,78 @@ Adds a screen manager method for changing the current layout to a new layout.
...
+We will add a new screen manager method for changing the current layout to a new layout. This also includes updating the template's night and day color schemes when the feature is available on systems supporting RPC spec 5.0.0 or newer. 
 
 ### iOS APIs
 ```objc
 @interface SDLScreenManager : NSObject
 
-/// Change the current layout to a new layout. You can get a list of SDL-defined layout strings in SDLPredefinedLayout.
-- (void)changeLayout:(NSString *)newLayout withCompletionHandler:(SDLScreenManagerUpdateCompletionHandler)handler;
+/// Change the current layout to a new layout, update the layout's night and day color schemes and update the text, graphics, buttons and template title.
+- (void)changeLayout:(SDLPredefinedLayout)newLayout templateUpdates:(nullable SDLTemplateUpdates *)templateUpdates withUpdateCompletionHandler:(SDLScreenManagerUpdateCompletionHandler)handler;
```
joeljfischer, line 24: This can't just use predefined layout or else custom templates won't work. See below requested updates.

joeljfischer, line 24:
```suggestion
- (void)changeLayout:(SDLTemplateLayout *)templateUpdates withCompletionHandler:(SDLScreenManagerUpdateCompletionHandler)handler;
```

proposals/0278-screenmanager-layout-management.md
```diff
@@ -11,33 +11,78 @@ Adds a screen manager method for changing the current layout to a new layout.
...
+
+@end
+
+@interface SDLTemplateUpdates : NSObject
+
```
joeljfischer, line 29: Add a property like:

```objc
@property (copy, nonatomic) NSString *layoutName;
```

proposals/0278-screenmanager-layout-management.md
```diff
@@ -11,33 +11,78 @@ Adds a screen manager method for changing the current layout to a new layout.
...
+@property (copy, nonatomic, nullable) NSString *mediaTrackTextField;
+@property (strong, nonatomic, nullable) SDLArtwork *primaryGraphic;
+@property (strong, nonatomic, nullable) SDLArtwork *secondaryGraphic;
+@property (copy, nonatomic) SDLTextAlignment textAlignment;
+@property (copy, nonatomic, nullable) SDLMetadataType textField1Type;
+@property (copy, nonatomic, nullable) SDLMetadataType textField2Type;
+@property (copy, nonatomic, nullable) SDLMetadataType textField3Type;
+@property (copy, nonatomic, nullable) SDLMetadataType textField4Type;
+@property (copy, nonatomic, nullable) NSString *title;
+@property (copy, nonatomic) NSArray<SDLSoftButtonObject *> *softButtonObjects;
```
joeljfischer, line 46: Add initializers for creating including the required layout name with either a string or predefined layout. It's probably fine to only have one initializer with the custom string and make someone who uses it set up the rest of it separately by setting properties. That way you don't have to duplicate every initializer. I would have something like:

1. Initializer with just String layout
2. Initializer with just PredefinedLayout
3. Initializer with PredefinedLayout, text fields 1-4, title, softButtonObjects
4. One with everything

proposals/0278-screenmanager-layout-management.md
```diff
@@ -11,33 +11,78 @@ Adds a screen manager method for changing the current layout to a new layout.
...
 ```
 
 ### Java APIs
 ```java
 abstract class BaseScreenManager extends BaseSubManager {
-    public void changeLayout(String newLayout, CompletionListener listener);
+    /*
+     * Change the current layout to a new layout, update the layout's night and day color schemes and update the text, graphics, buttons and template title.
+     */   
+    public void changeLayout (PredefinedLayout newLayout, TemplateUpdates templateUpdates, CompletionListener listener);
```
joeljfischer, line 56: See above

proposals/0278-screenmanager-layout-management.md
```diff
@@ -11,33 +11,78 @@ Adds a screen manager method for changing the current layout to a new layout.
...
+    List<SoftButtonObject> softButtonObjects;
 }
 ```
 
 ### JavaScript APIs
 The JavaScript APIs will be set up in a similar way to the Obj-C / Java APIs above. All changes will be at the discretion of the Project Maintainer. However larger changes that would impact the Objective-C code above (such as adding or removing a method) will require proposal revisions.
 
 ### Additional Implementation Notes
-1. When connected to systems running RPC <6.0, the screen manager should send `SetDisplayLayout` to change the template, and on systems running RPC >6.0, the screen manager should send `Show` to change the template.
+1. When connected to systems running RPC <6.0, the screen manager should send `SetDisplayLayout` to change the template. On receipt of  the `SetDisplayLayout` response, a `Show` will be sent by the existing Text and Graphic Manager and Soft Button Manager to update the text, graphic, and button information.
```
joeljfischer, line 83:
```suggestion
1. When connected to systems running RPC <6.0, the screen manager should send `SetDisplayLayout` to change the template. The graphics will be uploaded before sending the `SetDisplayLayout`. On receipt of the `SetDisplayLayout` response, a `Show` will be sent by the `ScreenManager` to update the text, graphic, and button information. The current implementation of the sub-managers will require the soft buttons to be sent after the rest of the template update, but that implementation detail could change in the future.
```

proposals/0278-screenmanager-layout-management.md
```diff
@@ -11,33 +11,78 @@ Adds a screen manager method for changing the current layout to a new layout.
...
 
 ### Additional Implementation Notes
-1. When connected to systems running RPC <6.0, the screen manager should send `SetDisplayLayout` to change the template, and on systems running RPC >6.0, the screen manager should send `Show` to change the template.
+1. When connected to systems running RPC <6.0, the screen manager should send `SetDisplayLayout` to change the template. On receipt of  the `SetDisplayLayout` response, a `Show` will be sent by the existing Text and Graphic Manager and Soft Button Manager to update the text, graphic, and button information. 
+1. When connected to systems running RPC >=6.0, the screen manager will send `Show` to change the template. The existing Text and Graphic Manager and Soft Button Managers will be used to send the updated `Show`. The Text and Graphic Manager will be used to upload the primary and secondary graphics and send the `Show` request. On receipt of the `Show` response, the Soft Button Manager will be used to send the updated soft buttons.
```
joeljfischer, line 84:
```suggestion
1. When connected to systems running RPC >=6.0, the screen manager will send `Show` to change the template. The `ScreenManager` will upload the graphics before sending the `Show`. The current implementation of the sub-managers will require the soft buttons to be sent after the rest of the template update, but that implementation detail could change in the future.
```

proposals/0278-screenmanager-layout-management.md
```diff
@@ -11,33 +11,96 @@ Adds a screen manager method for changing the current layout to a new layout.
...
+We will add a new screen manager method for changing the current layout to a new layout. This also includes updating the template's data immediately on v6.0+ and sequentially on <v6.0, as well as the template's color schemes on RPC v5.0+ systems.
 
 ### iOS APIs
 ```objc
 @interface SDLScreenManager : NSObject
 
-/// Change the current layout to a new layout. You can get a list of SDL-defined layout strings in SDLPredefinedLayout.
-- (void)changeLayout:(NSString *)newLayout withCompletionHandler:(SDLScreenManagerUpdateCompletionHandler)handler;
+/// Change the current layout to a new layout, update the layout's night and day color schemes and update the text, graphics, buttons and template title.
+- (void)changeLayout:(SDLTemplateLayout *)templateUpdates withCompletionHandler:(nullable SDLScreenManagerUpdateCompletionHandler)handler;
```
joeljfischer, line 24: I'm just thinking through this. If they provide a textField 1 & 2 only, would that clear everything else, or leave everything else alone (like happens currently in Screen Manager T&G updates)? I think that we'd want to do the second for now. If you agree, I think we should make that clear in the docs.

joeljfischer:
@NicoleYarroch I think that the proposed solution above needs to include all the new aspects to the changed proposal. Most people won't look at the changes, they'll just read the markdown text you have above. The markdown text hides a lot of the actual changes, and I think those need to be surfaced.

proposals/0278-screenmanager-layout-management.md
NicoleYarroch: You raised an interesting question. Since the values set for the text, graphics, buttons and template title persist between layout changes, if only the values set in the`templateUpdates` change then how can we keep track of whether the developer wants to leave the value unchanged or clear the value. Does setting `textField1` to `nil` mean we should clear the value or that we should not update the value at all? We could say that setting an empty string clears the `textField1` but I am not sure what the clear value would be for `primaryGraphic`, which has a data type of `SDLArtwork`.

joeljfischer: Honestly, the easiest way to do that would be to create a new (class) initializer for `SDLArtwork` like `[SDLArtwork clear]` or `nothing`. Otherwise, I don't know how we'd do that. You could also add a `SDLTemplateConfiguration` option to `clearUnsetData`.

Another option I just thought of that might be helpful is to remove all the fields from the `SDLTemplateLayout` class and then instead use `SDLTemplateConfiguration` which already exists. Then, if the dev batches the update, it'll all be merged together or set separately. If they don't batch, then it'll be set separately.

joeljfischer: I think I like that latter option quite a bit.

NicoleYarroch: I like your solution. Just to make sure we are on the same page, it sounds like you want to remove `SDLTemplateLayout` and just use `SDLTemplateConfiguration` like so:

```
- (void)changeLayout:(SDLTemplateConfiguration *)templateConfiguration withCompletionHandler:(nullable SDLScreenManagerUpdateCompletionHandler)handler;
```

The developer would do something like the following to batch the updates?
```
let templateConfiguration = SDLTemplateConfiguration(predefinedLayout: .graphicWithText)
templateConfiguration.dayColorScheme = SDLTemplateColorScheme(primaryColor: .red, secondaryColor: .blue, backgroundColor: .gray)

sdlManager.screenManager.beginUpdates()
sdlManager.screenManager.changeLayout(templateConfiguration) { error in
    // The template updated
}
sdlManager.screenManager.textField1 = "new Template set" // This will updat the current text in textField1
sdlManager.screenManager.textField2 = nil // This will clear the current text in textField2
sdlManager.screenManager.endUpdates { error in
    // Text, graphics, and buttons were updated
}
```

NicoleYarroch: If the `Show` request fails because the template is unsupported, then the entire `Show` request will be rejected (i.e. text, graphics and buttons will not be updated) however if the `SetDisplayLayout` request fails then the subsequent `Show` request will likely still succeed. Should it be the default to send another `Show` request without the template information if the request fails due to an invalid template (not even sure if we would be able to parse the error for this information) or should we let developers set the behavior?

joeljfischer: No, if the developer is changing the text / graphics alongside the layout change, I think it's reasonable to not set that text / graphics if the layout fails to change. If we're using the `SetDisplayLayout` and the layout change fails, we would not continue with the sequential `Show`. The same would happen with a `Show` with a layout change and the sequential soft buttons.

proposals/0278-screenmanager-layout-management.md
```diff
@@ -11,33 +11,38 @@ Adds a screen manager method for changing the current layout to a new layout.
...
+One big area that it is currently missing is the ability to manage changing templates, currently done with the `SetDisplayLayout` RPC (on RPC spec <v6.0) and `Show` RPC (on RPC spec v6.0+). This proposal adds a way to change the layout while taking care of backward compatibility concerns.
 
 ## Proposed solution
-We will add a new screen manager method for changing the current layout to a new layout.
+We will add a new screen manager method for changing the current layout to a new layout. This also includes updating the template's night and day color schemes on RPC v5.0+ systems. If the template update is batched with text, graphic updates on v6.0+ then all updates will be sent with one `Show` RPC (if buttons are also batched in the update, it will currently require a another `Show` request - this implementation detail could changein the future). On systems <v6.0, the `Show` request will be sent after  the `SetDisplayLayout`. If the layout update fails while batching, then the text, graphics, buttons or template title will also not be updated.
```
joeljfischer, line 17:
```suggestion
We will add a new screen manager method for changing the current layout to a new layout. This also includes updating the template's night and day color schemes on RPC v5.0+ systems. If the template update is batched with text and graphic updates on v6.0+ then all updates will be sent with one `Show` RPC (if buttons are also batched in the update, it will currently require a another `Show` request - this implementation detail could change in the future). On systems <v6.0, the `Show` request with the text and graphic updates will be sent after the `SetDisplayLayout`. If the layout update fails while batching, then the text, graphics, buttons or template title will also not be updated.
```

NicoleYarroch:

proposals/0278-screenmanager-layout-management.md
```diff
@@ -11,33 +11,38 @@ Adds a screen manager method for changing the current layout to a new layout.
...
+1. When connected to systems running RPC <v6.0, the screen manager will send a `SetDisplayLayout` to change the template. If the template update is batched with other updates, on receipt of a successful `SetDisplayLayout` response, the graphics will first be uploaded and then a `Show` request will be sent by the `ScreenManager` to update the text, graphic, and button information. If the `SetDisplayLayout` request fails, then the graphics and the `Show` request will not be sent in order to make sure that the behavior of the screen manager on RPC <v6.0 is consistent with RPC v6.0+. The current implementation of the sub-managers will require the soft buttons to be sent in a separate `Show` request after the rest of the template update (assuming the template update succeeds), but that implementation detail could change in the future.
```
theresalech, line 44:
```suggestion
1. When connected to systems running RPC <v6.0, the `ScreenManager` will send a `SetDisplayLayout` to change the template. If the template update is batched with other updates, on receipt of a successful `SetDisplayLayout` response, the graphics will first be uploaded and then a `Show` request will be sent by the `ScreenManager` to update the text, graphic, and button information. If the `SetDisplayLayout` request fails, then the graphics and the `Show` request will not be sent in order to make sure that the behavior of the screen manager on RPC <v6.0 is consistent with RPC v6.0+. The current implementation of the sub-managers will require the soft buttons to be sent in a separate `Show` request after the rest of the template update (assuming the template update succeeds), but that implementation detail could change in the future.
```

NicoleYarroch:
@theresalech I committed your suggestions. Ready for re-review.

PR #1018, opened 2020-05-26
Update process to defer proposal reviews that authors don't respond to

joeljfischer:
Add a statement to defer a proposal if the author has not responded to reviews for the full week their proposal is in review.

### DIFF
--- a/process.md
+++ b/process.md
@@ -114,7 +114,10 @@ In general, actions taken on proposals should be limited to one per week.  Pleas
 ### For Consideration
 The SDL Community is a global community, and so it can be expected that comments will be made on review issues at various times.  For this reason, we do not have a "cutoff" time for leaving comments on review issues.  While this may sometimes result in a proposal needing to be kept in review an additional week to allow the author or commenter time to respond, we appreciate the understanding of the community to allow such global discussions to take place.
 
-- If a proposal is in review and the author has not responded to the review for the full week, the proposal will be deferred until the author has acknowledged that they have the time to respond to reviews for the next week.
+- If a proposal is in review and comments have not been responded to by the author that were placed at least 3 full business days before the SDLC meeting, then the proposal will be deferred until the author has acknowledged that they have the time to respond to reviews for the next week.
+  - For example, the SDLC meeting is 6pm EDT on Tuesday. 3 full business days in U.S. time takes us back to 6pm EDT on Thursday. If comments were placed before then and were not responded to by the author, then the proposal will be deferred.
+  - If comments are placed after 6pm EDT on Thursday, the proposal would stay in review, however, if the author does not respond to those comments in that week of review, the proposal will then become deferred because more than 3 business days have passed.
+  - If the author is present on the SDLC meeting and confirms that they will respond to comments next week, the proposal may stay in review at the SDLC's discretion.
 
 ## Implementation Responsibility
 
--- a/process.md
+++ b/process.md
@@ -114,6 +114,8 @@ In general, actions taken on proposals should be limited to one per week.  Pleas
 ### For Consideration
 The SDL Community is a global community, and so it can be expected that comments will be made on review issues at various times.  For this reason, we do not have a "cutoff" time for leaving comments on review issues.  While this may sometimes result in a proposal needing to be kept in review an additional week to allow the author or commenter time to respond, we appreciate the understanding of the community to allow such global discussions to take place.
 
+- If a proposal is in review and the author has not responded to the review for the full week, the proposal will be deferred until the author has acknowledged that they have the time to respond to reviews for the next week.
+
 ## Implementation Responsibility
 
 The general rule of thumb is that the author of an SDL Evolution proposal is the person or company who intends to contribute code for a feature. SDLC Member contributions should encompass all projects impacted by the feature (including documentation).


kshala-ford:
@joeljfischer The update is fair allowing other proposals to get into review. Would it be possible to be more specific with the term "full week"? I also think that shutdown time, a golden week or similar may be considered as exceptions.

From the Ford group we would suggest another update. We have seen and we do see proposals of a certain age without comments until hours before the meeting. The suggestion is to define a priority to proposals currently in review. The lower the proposal number the higher the priority. The SDLC members, the proposal author and the PM should commit to support proposals of a higher priority and provide feedback in time to help to complete these proposals. Everyone should help to avoid proposals stay in review by providing feedback only hours before the meeting.

joeljfischer:
> The update is fair allowing other proposals to get into review. Would it be possible to be more specific with the term "full week"?

I think that's fair. What about an update that says that comments have been placed at least 3 full business days before the steering committee meeting (so in America, that would be a comment placed anytime Thursday, or if a comment was placed Friday - Tuesday, it would count for the _next_ steering committee meeting).

> I also think that shutdown time, a golden week or similar may be considered as exceptions.

I disagree. I think that the member should have informed the SDLC that they would not be capable of defending their proposal that week and so it should not have been in review. If they do not do so and don't respond to comments, I think it's fair to pull it out. However, I would also say that if the member is on the call and can confirm that they will respond for the next week, the SDLC can keep that proposal in review.

> From the Ford group we would suggest another update.

I think this is out of scope for this PR, and would encourage you to open a PR with specifics if you'd like to see this change.

kshala-ford:
Let me try to rephrase if I fully understand. If the author hasn't replied to comments which were placed until two days after the last SC meeting (that would be Wednesday and Thursday) the proposal should be subject for deferral? And the proposal should stay in review for comments placed Friday-Tuesday?

> I disagree. I think that the member should have informed the SDLC that they would not be capable of defending their proposal that week and so it should not have been in review.

Some proposals are in review for weeks or months. Sure the SDLC member should inform the PM that the member may not be available for a week. First this reduces the number of possible votes and second I don't know if the proposal can be deferred for this specific week only. The member should instead be encouraged to provide feedback on the "Friday" before the shutdown but also inform about the absence at the next SC meeting.

joeljfischer:
> Let me try to rephrase if I fully understand.

I've updated the proposal to hopefully make it clearer.

> Some proposals are in review for weeks or months. Sure the SDLC member should inform the PM that the member may not be available for a week. First this reduces the number of possible votes and second I don't know if the proposal can be deferred for this specific week only. The member should instead be encouraged to provide feedback on the "Friday" before the shutdown but also inform about the absence at the next SC meeting.

If a member doesn't respond to comments and isn't present for the next SDLC meeting, the proposal is deferred, even if it's been in the process for weeks and months. It can be pulled back in at the SDLC's discretion when the member confirms they are available to defend their proposal. I don't believe it is fair to the rest of the SDLC members who have proposals waiting for review for a member to take a week off while their proposal is in review without notice. 

The member should be encouraged to provide their feedback on Friday, but if the member isn't available for the SDLC meeting, I believe their proposals should be deferred that week. The proposal can be pulled back in when there's an available slot. That's outside the scope of this proposal though and can be done at the SDLC's discretion.

theresalech:
Including action item from 2020-06-02 Steering Committee meeting minutes here for visibility:

>Action Item: author to update PR per this comment, and merge PR into `master`.

kshala-ford:
> That's outside the scope of this proposal though and can be done at the SDLC's discretion.

Your change affects proposals in review so it absolutely is in scope. I don't think Ford can accept this statement given the number of authors of different proposals. Also I don't think your proposal is acceptable given the fact that the SDLC works globally. European members cannot defend their proposals unless you want them to dial in at midnight? 

Deferring a proposal on an SC meeting delays the progress of the proposal significantly. For instance the author can inform the PM about possible absence in advance and could reply to feedback on the last "Friday". If the proposal is deferred for a week it will not be possible to work on the proposal for at least three full days in the following week allowing no one to provide feedback. Given the risk that the proposal could not get back to review if no slots are available.

We digress from what motivated you to propose the change. The proposal author should have the chance to defend the proposal in the GitHub issue. It should not be required for the author to join the meeting.

joeljfischer:
First, this proposed change was accepted by the SDLC at the last meeting. If you'd like to change it, you will have to have a new PR approved by the SDLC. However, I think there are many misconceptions that I can clarify below.

> Also I don't think your proposal is acceptable given the fact that the SDLC works globally. European members cannot defend their proposals unless you want them to dial in at midnight?

The "author" includes the company on the call. If it's a European member that made the proposal, they either need to be active in defending their review, or, if they're not, and are on the call, they can confirm to the SDLC that they can defend the proposal the next week and it can stay in review at the SDLC's discretion. If it is a global member, any member of that company could confirm that they will defend the proposal the next week. Otherwise, normal processes can take place and they can confirm to the maintainer another way (e.g. email) and it can be pulled in at the next meeting at the SDLC's discretion.

> Deferring a proposal on an SC meeting delays the progress of the proposal significantly. For instance the author can inform the PM about possible absence in advance and could reply to feedback on the last "Friday". If the proposal is deferred for a week it will not be possible to work on the proposal for at least three full days in the following week allowing no one to provide feedback.

If it is that important and urgent to the author, they should not have left it without commenting for a week. However, if they reply on that last "Friday," this change will not take effect, so I'm not sure what the issue is.

> Given the risk that the proposal could not get back to review if no slots are available.

There is almost always at least one open slot available.

> The proposal author should have the chance to defend the proposal in the GitHub issue. It should not be required for the author to join the meeting.

They can, if you read the change, they can comment on the issue and keep it active. If they comment on the issue, this rule change doesn't come into effect, so I'm not sure what the issue is here.

PR #1033, opened 2020-06-09
Update process to prioritize older proposals

kshala-ford:
Add a statement that describes how older proposals should be handled of higher priority. Everyone should try to provide feedback in time and avoid older proposals to stay idle.


kshala-ford:
@theresalech This PR is ready for review. Please add this proposed process change to the agenda. There is a noticeable amount of proposals in the last couple of months where comments are made in last minute. This change should motivate everyone to prioritize older proposals to get finalized and allow more room for new and future proposals

process.md
```diff
@@ -111,6 +111,10 @@ In general, actions taken on proposals should be limited to one per week.  Pleas
 - If the Steering Committee votes to return a proposal for revisions, bringing the revised proposal back into review cannot happen until the following week.  This allows for adequate time to ensure the Steering Committee's agreed upon revisions have been made.  
 - A proposal cannot be brought into review and voted upon in the same week, as Steering Committee representatives would not have sufficient time to review the proposal prior to voting.
 
+### Priority
+
+For many different reasons it can be possible that proposals remain in review for a couple of iterations. While it is important to improve the proposals to contain a complete and mature solution before acceptance these proposals occupy slots for other proposals. Therefore older proposals should be treated with a higher priority. The SDLC members, the proposal author and the PM should commit to support proposals of a higher priority and provide feedback in time to help to complete these proposals. The idle time of these proposals should be avoided as much as possible. Everyone should help to complete proposal and allow new proposals to get into review.
```
theresalech, line 116:
```suggestion
For many different reasons it can be possible that a proposal remain in review for a couple of iterations. While it is important to improve the proposal to contain a complete and mature solution before acceptance, each proposal occupies a potential slot for other proposals awaiting review. Therefore older proposals should be treated with a higher priority. The SDLC members, the proposal author, and the PM should commit to supporting proposals of a higher priority and providing feedback in time to help to reach a decision on these proposals. The idle time of these proposals should be avoided as much as possible. Everyone should help to reach a decision on a proposal and allow new proposals to enter review.
```

kshala-ford:
Thank you @theresalech I updated the PR as you suggested.

theresalech:
@kshala-ford, thanks!  I'll add this to the next Steering Committee meeting agenda for discussion.

joeygrover:
I believe the general idea behind this change has merit and that the goal should always be to get proposals through the process as efficiently as possible. I also believe that it is everyone's intention to move proposals through and not deliberately ignoring proposals that are in need of attention. That would beg the question what is this added language providing that isn't already being done? I'm hesitant to put in such language for this reason and a few more. 

The wording here is more a description of "should" and not of "must". This leads to a process that is up for interpretation and we should avoid such situations. I don't believe there's anything within this addition that can be regulated. For example, what would be the consequences for a member commenting on a newer proposal but not an older one which should (based on this addition) have higher priority?  Also, should we penalize a member for commenting on a newer proposal but not an older one, when there are other considerations in play (such as member-specific priority or applicability)? The goal behind this document was to provide specific processes that can be followed and not be bent into different meanings and outcomes. 

Next, if the previous item was fixed to be more concrete we would run into the issue of practicality when it comes to the scope of the proposals in review. It is much easier to review smaller proposals that have limited scope compared to large features that require much debate. Therefore, even if a proposal has been in the process for a long time it becomes unwise to simply require it to be reviewed before the more easily reviewed proposals for the only purpose to not take up a review slot. During some review periods, such proposals that are more narrowly scoped are more likely to be reviewed and an outcome provided than those that have been debated to death. 

These situations are not always as easy to distinguish and are even harder to put into concrete processes. For these reasons I would recommend not including this language so that we can avoid vague processes as well as ones that would actually make the problem worse that it would try to solve.

theresalech:
Closing based on Steering Committee discussion and comment above.

PR #1081, opened 2020-10-06
SDL 0307 - Get Input Screen Revisions

MichaelCrimando:
This proposal adds an Form screen for apps. This enables a way for the user to sign in, put in their credit card information, put in their address and more. This is particularly focused on embedded/cloud apps because the app lives in the car and the user will still need a way to enter information without having the ability to do so on their phone.


MichaelCrimando:
FYI this has been de-prioritized.

PR #1093, opened 2020-11-17
App Library Remote Control Manager

joeljfischer:
This proposal adds a remote control manager to the app libraries which manages everything related to the remote control systems on head units and allows apps to more conveniently retrieve and set remote control data.


joeljfischer:
@pelsila This proposal is ready for review

proposals/nnnn-app-lib-remote-control-manager.md
```diff
@@ -0,0 +1,315 @@
...
+public void setUserSeatLocation(Grid location, RemoteControlCompletionListener completionListener)
+```
+
+##### JavaScript Suite
+```js
+async setUserSeatLocation(location)
+```
+
+#### Getting User Consent
+On RPC 6.0+ systems, an app may be required to request access to a remote control module from the user, but it should only do when it actually needs that access. The developer will be able to request user consent manually for a module or modules whenever it makes sense within their app's flow.
```
pelsila, line 68:
```suggestion
On RPC 6.0+ systems, an app may be required to request access to a remote control module from the user, but only when it actually needs that access. The developer will be able to request user consent manually for a module or modules whenever it makes sense within their app's flow.
```

pelsila (changes requested):
@joeljfischer Just one minor suggestion.  Let me know if you have any questions.  Thank you

joeljfischer:
@pelsila Committed the suggestion, thank you for the review.

pelsila (approved):

PR #1135, opened 2021-04-29
Handle Scenario Where no Valid Cert is Available

iCollin:
This proposal defines how Core should respond to a request to start a protected service when it does not have valid certificates to complete a handshake.


iCollin:
@theresalech Ready for review

theresalech (approved):

PR #1143, opened 2021-05-13
Strict Versioning for Outgoing Core Messages

jacobkeeler:
The main purpose of this proposal is to verify that SDL Core follows the Mobile API spec properly when communicating with older apps. Currently, due to changes in the Mobile API spec over time, there are cases where SDL Core can send a message to an older app which isn't valid for its API version.


jacobkeeler:
@theresalech Ready for review

pelsila (approved):

Issue #1144, opened 2021-05-19
[Accepted] SDL 0336 - Strict Versioning for Outgoing Core Messages


theresalech:
Hello SDL community,

The review of the revised proposal "SDL 0336 - Strict Versioning for Outgoing Core Messages" begins now and runs through June 29, 2021. Previous reviews of this proposal took place May 19 - June 1, 2021 and June 9 - June 15, 2021. The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0336-strict-versioning-for-outgoing-core-messages.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/1144

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Theresa Lech

Program Manager - Livio
theresa@livio.io

joeljfischer:
1\. I think that when a parameter or structure is dropped on Core, the message should be change from SUCCESS to WARNINGS and an info string should be added mentioning what was dropped and why (like "gps parameter was dropped due to a RPC version mismatch, please update the app library to fix this issue").

joeygrover:
1\. Would this be breaking for older apps? Since they will expect the `success` flag to be set to true and for the response code to equal `SUCCESS`. Changing that to a `WARNINGS` result code implies something went wrong even though every piece of the message otherwise aligns with the RPC spec version that was negotiated. Adding that info or similar to the `info` string makes sense, but I disagree with changing the result code as I believe it is a bit misleading and a potential breaking change for an app developer.

joeljfischer:
[...]

PR #1155, opened 2021-06-16
Revise SDL-0336 Strict Versioning for Outgoing Core Messages

jacobkeeler:
Revised proposal based on review comments (see https://github.com/smartdevicelink/sdl_evolution/issues/1144#issuecomment-862359945)

## Updates

- Edited sanitation process to only remove non-mandatory values
- Added note for integrators to update minimum version when sending data that does not work with older apps
- Removed details for generating replacement values
- Updated examples based on new rules


jacobkeeler:
@theresalech Ready for review

theresalech (approved):

Issue #1013, opened 2020-05-20
[Returned for Revisions] SDL 0303 - SDL Audio I/O Manager


theresalech:
Hello SDL community,

The review of "SDL 0303 - SDL Audio I/O Manager" begins now and runs through September 22, 2020. Previous reviews of this proposal took place May 19 - 26, 2020 and June 17 - July 14, 2020. The proposal is available here:

https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0303-audio-io-manager.md

Reviews are an important part of the SDL evolution process. All reviews should be sent to the associated Github issue at:

https://github.com/smartdevicelink/sdl_evolution/issues/1013

What goes into a review?

The goal of the review process is to improve the proposal under review through constructive criticism and, eventually, determine the direction of SDL. When writing your review, here are some questions you might want to answer in your review:

* Is the problem being addressed significant enough to warrant a change to SDL?
* Does this proposal fit well with the feel and direction of SDL?
* If you have used competitors with a similar feature, how do you feel that this proposal compares to those?
* How much effort did you put into your review? A glance, a quick reading, or an in-depth study?
Please state explicitly whether you believe that the proposal should be accepted into SDL.

More information about the SDL evolution process is available at

https://github.com/smartdevicelink/sdl_evolution/blob/master/process.md

Thank you,
Theresa Lech

Program Manager - Livio
theresa@livio.io

joeljfischer:
1\. This proposal is in conflict with the accepted proposal [SDL-0295](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0295-screen-manager-audio-pass-thru.md). I would recommend that this proposal be rejected in favor of a revisions proposal to that one (or a new proposal) that adds some of the features from this proposal (namely, pausing audio streaming output when input is requested and audio level normalization) to the structure of that proposal.

2\. I don't believe that having the entirety of the code be in the proposal is a good idea. The maintainers must effectively attempt to integrate the code and provide a code review in the proposal comments for the author to update. Such a process would take a _very_ long time and is, in my view, a waste of everybody's time.

3\. I don't see how we could accept this proposal without corresponding Java and JavaScript impact. Adding features to iOS only that can and should be on all app libraries undos a lot of work that the PM has gone through to make the libraries more aligned.

Given those recommendations, I will hold off reviewing the rest of the proposal until the SDLC has decided how to proceed with this proposal.

theresalech:
[...]

PR #1165, opened 2021-07-15
Revise SDL 0303 - SDL Audio I/O Manager

ghost:
This proposal is about integrating a new manager to the sdl_ios library that allows reading audio data from the car microphone and sending audio to the car speaker.

This PR includes the accepted revisions in https://github.com/smartdevicelink/sdl_evolution/issues/1013#issuecomment-682101300 and a minor correction to the state diagram and code. Also includes a note to proposal 0297 to respect the accepted change sending audio in chunks.


Martel5011 (changes requested):
`

PR #1173, opened 2021-08-05
Revise SDL 0338 - Remove Unused Files From SDL Core

JackLivio:
# Introduction

This is a revision to an accepted proposal. The revision is to keep two files that were found to be used in the project.

# Motivation

During the implementation of this proposal it was found that system.cc and system.h are still used by SDL Core and should not be removed from the project.

# Proposed Solution

Update the proposal to not include system.cc and system.h in the list of files to remove.

# Potential Downsides

The author does not see any potential downsides that weren't already addressed in the original proposal.

# Impact On Existing Code

system.cc and system.h will be kept in the SDL Core project.

# Alternatives Considered

The alternative would be to completely revert the original proposal.

### DIFF
--- a/proposals/0338-remove-unused-files-from-sdl-core.md
+++ b/proposals/0338-remove-unused-files-from-sdl-core.md
@@ -17,8 +17,6 @@ In an effort to trim the code base of SDL Core the author would like to remove s
 
 Remove the following files:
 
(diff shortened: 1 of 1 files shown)


JackLivio:
@theresalech Ready for review

theresalech (approved):

theresalech (approved):

PR #1180, opened 2021-12-06
Android 12 Issues

RHenigan:
With almost every Android version, Google updates the OS to better distribute the device resources among the running apps to enhance the performance and optimize battery life. However, sometimes the new enhancements come with new restrictions on how apps can use the resources on the device. Android 12 is one of the major Android updates that heavily modify how apps can run in the background and use device resources. This affects how the `SdlRouterService` starts and communicates with other SDL apps and also how the `SdlService` for SDL apps is started. This proposal is to address the issues that have been introduced as a result of Android 12.

### DIFF
--- a/proposals/0345-android-12-issues.md
+++ b/proposals/0345-android-12-issues.md
@@ -1,6 +1,6 @@
 # Android 12 Issues
 
-* Proposal: [SDL-0345](0000-android-12-issues.md)
+* Proposal: [SDL-NNNN](nnnn-android-12-issues.md)
 * Author: [Robert Henigan](https://github.com/RHenigan)
 * Status: **Awaiting review**
 * Impacted Platforms: [Java Suite]
@@ -28,11 +28,11 @@ Android 12 adds multiple restrictions on how apps can work in the background and
 ### Foreground Services
 With the new restrictions around starting foreground services from the background we will need to meet one of the exceptions that Android 12 would allow to start a foreground service from the background or we would need to start the foreground service from a foreground context.
 
-PendingIntents allows us to send an intent on behalf of another app using the context and permissions of the application that created the PendingIntent. Since the `SdlRouterService` will run in the foreground, the `SdlRouterService` could create a PendingIntent to start an apps  `SdlService`. Since the `SdlRouterService` is creating the PendingIntent, the intent would be sent from the `SdlRouterServices` Context.
+`PendingIntents` allows us to send an intent on behalf of another app using the context and permissions of the application that created the `PendingIntent`. Since the `SdlRouterService` will run in the foreground, the `SdlRouterService` could create a `PendingIntent` to start an app's `SdlService`. Since the `SdlRouterService` is creating the `PendingIntent`, the intent would be sent from the `SdlRouterService`'s context.
 
-The PendingIntent in the `SdlRouterService` can be given to the `SdlBroadcastReceiver`. At this point the `SdlReceiver` will be able to send the PendingIntent with and updated Intent where the developer specifies their unique `SdlService` class. By calling `PendingIntent.send()` we are starting a new foreground service of the unique `SdlService` class from the context of the `SdlRouterService`. Since the `SdlRouterService` is running in the foreground we are not longer trying to start a foreground service from the background.
+The `PendingIntent` in the `SdlRouterService` can be given to the `SdlBroadcastReceiver`. At this point the `SdlReceiver` will be able to send the `PendingIntent` with and updated Intent where the developer specifies their unique `SdlService` class. By calling `PendingIntent.send()` we are starting a new foreground service of the unique `SdlService` class from the context of the `SdlRouterService`. Since the `SdlRouterService` is running in the foreground we are no longer trying to start a foreground service from the background.
 
-This will require that the exported flag be set to true in the manifest for the apps `SdlService` as an external apps `SdlRouterService` would be the one starting the service.
+This will require that the exported flag be set to `true` in the manifest for the app's `SdlService`, as an external app's `SdlRouterService` would be the one starting the service.
 
 #### Library Changes
 ##### AndroidTools.java
@@ -93,9 +93,9 @@ public void onSdlEnabled(Context context, Intent intent) {
 ~~~
 
 ### Bluetooth Runtime Permissions
-With the new required Bluetooth Runtime permissions we will need developers to include the new permissions in the `AndroidManifest.xml` file. 
+With the new required Bluetooth runtime permissions we will need developers to include the new permissions in their `AndroidManifest.xml` file. 
 
-#### Developer Changes
+#### App Developer Changes
 ##### AndroidManifest.xml
 ~~~ xml
 <uses-permission android:name="android.permission.BLUETOOTH_CONNECT"
@@ -107,7 +107,7 @@ With the new required Bluetooth Runtime permissions we will need developers to i
 
 The developers will also need to request these permissions from the user as they are runtime permissions. If the user does not grant these permissions for a given application then that application will not receive any Intent with the `ACL_CONNECTED` action in the `SdlBroadcastReceiver`.
 
-In the event that the user denies bluetooth permissions from the application this complicates the use cases surrounding the `SdlBroadcastReceiver` and `SdlRouterService`. If the permissions are denied for the application then the apps `SdlBroadcastReceiver` will not receive any `ACL_CONNECTED` broadcasts and therefore will not know to start its own `SdlRouterService` when the device connects over bluetooth. Other SDL Apps on the device will see that they were not the designated app to start the `SdlRouterService` and will simply do nothing meaning no `SdlRouterService` will be started even if most of the SDL Apps do have the bluetooth permissions. To correct this in the `SdlBroadcastReceiver` we can see if the designated app has the bluetooth permissions. If they do we will let them start their `SdlRouterService` as normal. If they do not we will search the list of SDL Apps for the first app with the bluetooth permissions, that app will now be considered the designated app to start its `SdlRouterService`. The designated App will then try to start its own `SdlRouterService`.
+In the event that the user denies Bluetooth permissions from the application, this complicates the use cases surrounding the `SdlBroadcastReceiver` and `SdlRouterService`. If the permissions are denied for the application then the app's `SdlBroadcastReceiver` will not receive any `ACL_CONNECTED` broadcasts and therefore will not know to start its own `SdlRouterService` when the device connects over Bluetooth. Other SDL apps on the device will see that they were not the designated app to start the `SdlRouterService` and will simply do nothing, meaning no `SdlRouterService` will be started even if most of the SDL apps do have the Bluetooth permissions. To correct this, in the `SdlBroadcastReceiver` we can see if the designated app has the Bluetooth permissions. If they do, we will let them start their `SdlRouterService` as normal. If they do not, we will search the list of SDL apps for the first app with the Bluetooth permissions; that app will now be considered the designated app to start its `SdlRouterService`. The designated app will then try to start its own `SdlRouterService`.
 
 #### Library Changes
 ##### SdlBroadCastReceiver.java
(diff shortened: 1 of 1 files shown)


RHenigan:
@theresalech This proposal is ready for review

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+
+PendingIntents allows us to send an intent on behalf of another app using the context and permissions of the application that created the PendingIntent. Since the `SdlRouterService` will run in the foreground, the `SdlRouterService` could create a PendingIntent to start an apps  `SdlService`. Since the `SdlRouterService` is creating the PendingIntent, the intent would be sent from the `SdlRouterServices` Context.
+
+The PendingIntent in the `SdlRouterService` can be given to the `SdlBroadcastReceiver`. At this point the `SdlReceiver` will be able to send the PendingIntent with and updated Intent where the developer specifies their unique `SdlService` class. By calling `PendingIntent.send()` we are starting a new foreground service of the unique `SdlService` class from the context of the `SdlRouterService`. Since the `SdlRouterService` is running in the foreground we are not longer trying to start a foreground service from the background.
```
theresalech, line 33:
```suggestion
The `PendingIntent` in the `SdlRouterService` can be given to the `SdlBroadcastReceiver`. At this point the `SdlReceiver` will be able to send the `PendingIntent` with and updated Intent where the developer specifies their unique `SdlService` class. By calling `PendingIntent.send()` we are starting a new foreground service of the unique `SdlService` class from the context of the `SdlRouterService`. Since the `SdlRouterService` is running in the foreground we are no longer trying to start a foreground service from the background.
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+
+In the event that the user denies bluetooth permissions from the application this complicates the use cases surrounding the `SdlBroadcastReceiver` and `SdlRouterService`. If the permissions are denied for the application then the apps `SdlBroadcastReceiver` will not receive any `ACL_CONNECTED` broadcasts and therefore will not know to start its own `SdlRouterService` when the device connects over bluetooth. Other SDL Apps on the device will see that they were not the designated app to start the `SdlRouterService` and will simply do nothing meaning no `SdlRouterService` will be started even if most of the SDL Apps do have the bluetooth permissions. To correct this in the `SdlBroadcastReceiver` we can see if the designated app has the bluetooth permissions. If they do we will let them start their `SdlRouterService` as normal. If they do not we will search the list of SDL Apps for the first app with the bluetooth permissions, that app will now be considered the designated app to start its `SdlRouterService`. The designated App will then try to start its own `SdlRouterService`.
```
theresalech, line 110:
```suggestion
In the event that the user denies Bluetooth permissions from the application, this complicates the use cases surrounding the `SdlBroadcastReceiver` and `SdlRouterService`. If the permissions are denied for the application then the app's `SdlBroadcastReceiver` will not receive any `ACL_CONNECTED` broadcasts and therefore will not know to start its own `SdlRouterService` when the device connects over Bluetooth. Other SDL apps on the device will see that they were not the designated app to start the `SdlRouterService` and will simply do nothing, meaning no `SdlRouterService` will be started even if most of the SDL apps do have the Bluetooth permissions. To correct this, in the `SdlBroadcastReceiver` we can see if the designated app has the Bluetooth permissions. If they do, we will let them start their `SdlRouterService` as normal. If they do not, we will search the list of SDL apps for the first app with the Bluetooth permissions; that app will now be considered the designated app to start its `SdlRouterService`. The designated app will then try to start its own `SdlRouterService`.
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+    PendingIntent pendingIntent = PendingIntent.getActivity(this, 0, intent, flag);
+    //...
+}
+~~~
+
+### Service Notification Delays
+
+Starting in Android 12 when a service tries to present a notification, Android may delay showing the notification for up to 10 seconds. This is to try to allow the service to complete before the notification is presented. If we have connected to a system before we may want to present the `SdlRouterService` notifications immediately. App developers may also want to display the notifications related to the `SdlService` immediately. This can be achieved by setting the foregroundServiceBehavior flag to `Notification.FOREGROUND_SERVICE_IMMEDIATE`.
+
+In the event this is the first time the `SdlRouterService` is trying to connect to a system or if the `SdlRouterService` has seen but not connected with the system we will not use this flag and allow Android to hide the notification.
```
theresalech, line 252:
```suggestion
In the event this is the first time the `SdlRouterService` is trying to connect to a system or if the `SdlRouterService` has seen but not connected with the system, we will not use this flag and instead allow Android to hide the notification.
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+    }
+    //...
+}
+~~~
+
+## Potential downsides
+
+### Foreground Services
+
+Using `PendingIntents` to start the apps `SdlService` from the `SdlRouterService` creates a potential security risk as the apps `SdlService` will now be required to have the `android:exported` attribute set to true. This could expose apps to have their `SdlService` be started by apps that are not SDL certified.
```
theresalech, line 272:
```suggestion
Using `PendingIntents` to start the app's `SdlService` from the `SdlRouterService` creates a potential security risk as the app's `SdlService` will now be required to have the `android:exported` attribute set to `true`. This could expose apps to have their `SdlService` be started by apps that are not SDLC certified. 
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Bluetooth Runtime Permissions
+
+With the new runtime permissions, users will have to grant bluetooth permissions to SDL enabled apps at runtime. This means that if the user denies permissions for a specific app, the app will not know when the device connects to the head unit over bluetooth nor will it be able to start its `bluetoothTransport` in the `SdlRouterService`. However, apps that have their bluetooth permissions denied will still be able to bind to another app's `SdlRouterService` and will still be able to start up a `SdlRouterService` for a USB connection only. If there are not any apps on the device with bluetooth permissions granted then the `SdlRouterService` would never be started when connecting over bluetooth.
+
+## Impact on existing code
+
+### Foreground Services
+
+The `SdlRouterService` currently sends an intent out for each apps `SdlBroadcastReceiver` which notifies the apps to start the `SdlService`. If the app targets Android 12 and tries to start the `SdlService` from the background like this the app will crash. We will need to add a PendingIntent as an extra to this intent to be used by the `SdlReceiver`
```
theresalech, line 282:
```suggestion
The `SdlRouterService` currently sends an intent out for each app's `SdlBroadcastReceiver` which notifies the apps to start the `SdlService`. If the app targets Android 12 and tries to start the `SdlService` from the background like this, the app will crash. We will need to add a `PendingIntent` as an extra to this intent to be used by the `SdlReceiver`.
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+These changes will require app developers to change how they implement the `SdlReceiver.onSdlEnabled` method to check the incoming intent for a pendingIntent and send that pendingIntent with the appropriate class name.
+This implementation will also require the `SdlService` to have the `android:exported` attribute set to `true` in the `AndroidManifest.xml`.
+
+### Bluetooth Runtime Permissions
+
+The bluetooth logic in the library will now require these 2 new runtime permissions. With these permissions granted there is not any issue with the bluetooth logic but we now need to account for the use cases where a user has denied permissions. Specifically without these permissions the `SdlBroadcastReceiver` will never receive Intents with the `ACL_CONNECTED` action and the `SdlRouterService` will not be able to start the `bluetoothTransport`.
```
theresalech, line 288:
```suggestion
The Bluetooth logic in the library will now require these 2 new runtime permissions. With these permissions granted, there is not any issue with the Bluetooth logic, but we now need to account for the use cases where a user has denied permissions. Specifically, without these permissions the `SdlBroadcastReceiver` will never receive Intents with the `ACL_CONNECTED` action and the `SdlRouterService` will not be able to start the `bluetoothTransport`.
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+
+### AndroidManifest Exported Flag
+
+These attributes were already defined where required. The `SdlService` may also require this flag to be set but not as a direct result of Android 12 and rather because of the proposed Solution of starting the `SdlService` from the `SdlRouterService`.
```
theresalech, line 292:
```suggestion
These attributes were already defined where required. The `SdlService` may also require this flag to be set but not as a direct result of Android 12 and rather because of the proposed solution of starting the `SdlService` from the `SdlRouterService`.
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+
+### PendingIntent Mutable Flag
+
+The `SdlRouterService` will crash if the the app targets Android 12 and does not have this flag defined on the PendingIntent used for `SdlRouterService` notifications.
```
theresalech, line 296:
```suggestion
The `SdlRouterService` will crash if the the app targets Android 12 and does not have this flag defined on the `PendingIntent` used for `SdlRouterService` notifications.
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Service Notification Delays
+
+These notifications may be delayed by Android by up to 10 seconds but we can modify the notification builder to display them immediately as this is the desired behaviour in some cases.
```
theresalech, line 300:
```suggestion
These notifications may be delayed by Android by up to 10 seconds, but we can modify the notification builder to display them immediately as this is the desired behavior in some cases.
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+
+## Alternatives considered
+
+### Foreground Services
+
+Alternatives for the Foreground Service restrictions are limited. We either need to start the `SdlService` from a foreground context or the conditions need to meet one of the exceptions listed by Google. These conditions include:
+
+* Starting the `SdlService` from an Activity.
```
theresalech, line 308:
```suggestion
* Starting the `SdlService` from an Activity
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+* Starting the `SdlService` from user interaction with a notification
+* Requesting the user ignores battery optimizations for each SDL Application
```
theresalech, line 310:
```suggestion
* Requesting the user ignores battery optimizations for each SDL application
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+
+These options would either require an Activity to be launched for each SDL app, the user to interact with a notification for each SDL app, or for the user to choose battery optimization options for each app which then creates a situation where the user dictates if the given app will have its `SdlService` start when the `SdlRouterService` connects.
+
+If we decide to use `PendingIntents` and export the `SdlService` we could also create a custom SDL `<permission>` to be used by apps as a requirement to be able to start the `SdlService` but nothing is stopping developers from listing that custom permission in their `AndroidManifest.xml`.
```
theresalech, line 314:
```suggestion
If we decide to use `PendingIntents` and export the `SdlService` we could also create a custom SDL `<permission>` to be used by apps as a requirement to be able to start the `SdlService`, but nothing is stopping developers from listing that custom permission in their `AndroidManifest.xml`.   
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+
+If we decide to use `PendingIntents` and export the `SdlService` we could also create a custom SDL `<permission>` to be used by apps as a requirement to be able to start the `SdlService` but nothing is stopping developers from listing that custom permission in their `AndroidManifest.xml`.   
+
+### Bluetooth Runtime Permissions
+
+In the use case of the device only being connected over USB we could try to start the `SdlRouterService` of an application that does have Bluetooth Permissions to ensure we select a `SdlRouterService` that can start the `bluetoothTransport`. The problem with this solution is when we choose to start a `SdlRouterService` that is different than the one started by the `USBAccessoryAttachmentActivity`, the `USBTransferProvider` will try to bind to the wrong `RouterService`. This will require a larger refactor of the `USBTransferProvider`.
```
theresalech, line 318:
```suggestion
In the use case of the device only being connected over USB, we could try to start the `SdlRouterService` of an application that does have Bluetooth permissions to ensure we select a `SdlRouterService` that can start the `bluetoothTransport`. The problem with this solution is when we choose to start a `SdlRouterService` that is different than the one started by the `USBAccessoryAttachmentActivity`, the `USBTransferProvider` will try to bind to the wrong `RouterService`. This will require a larger refactor of the `USBTransferProvider`.
```

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+            context.startService(intent);
+        }
+    }
+}
+~~~
+
+### Bluetooth Runtime Permissions
+With the new required Bluetooth Runtime permissions we will need developers to include the new permissions in the `AndroidManifest.xml` file. 
+
+#### Developer Changes
```
theresalech, line 98:
```suggestion
#### App Developer Changes
```

Does it make sense to move this section to after "Library Changes" to keep the same order as the previous section?

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+
+If a phone connects to the head unit over USB we can still start the `SdlRouterService` but if the designated app does not have Bluetooth Permissions the `bluetoothTransport` will still not be able to start. This could cause confusion for the user if they expect the `SdlRouterService` to connect over USB and Bluetooth but the `SdlRouterService` will only connect over USB. The suggested solution is to present a notification to the user reminding them to enable Bluetooth Permissions. By clicking on the notification the user will be directed to the apps settings page to grant those permissions. Meanwhile the `SdlRouterService` will wait to initialize the `bluetoothTransport` and will continuously check the permission status. Once the permissions are granted the `bluetoothTransport` will be started.
+
+#### Library Changes
```
theresalech, line 172: This header is already present on line 112. I think it can be deleted here.

proposals/0345-android-12-issues.md
```diff
@@ -0,0 +1,322 @@
...
+
+    return true;
+}
+~~~
+
+### AndroidManifest Exported Flag
+
+Starting in Android 12 any activities, services, or broadcast receivers that use intent filters will need to explicitly declare the `android:exported` attribute for the given app components. The `SdlRouterService` and `SdlReceiver` should already have the exported attribute defined and set to true. But The `USBAccessoryAttachmentActivity` will now also require this attribute to be set. Any activity that had an `intent-filter` would have a default exported value of true. Now we need to explicitly set it.
+
+#### Developer Changes
```
theresalech, line 218:
```suggestion
#### App Developer Changes
```

theresalech (changes requested):
Hi @RHenigan , I've completed my review! Please let me know if you have any questions or concerns about my comments. Thank you!

proposals/0345-android-12-issues.md
RHenigan: My reasoning for this order is that this is directly related to the topic in the line above and it makes sense to start with the changes to the manifest before the impact the permissions has on the library.

proposals/0345-android-12-issues.md
RHenigan: removed

RHenigan:
@theresalech ready for re-review

theresalech (approved):

